PR #9, opened 2022-04-13
PoS model - remove pending

angbrav:
This PR:
- Removes pending state completely and adapt transactions consequently
- Changes is_validator function
- Defines slash_rate function
- Removes comments already addressed
- Fixes some minor typos

- [ ]  Safety checks on validators state is only done at the offset. We should check that this is OK. For instance, the is_validator function only checks that validator.state == candidate at cur_epoch+offset. 
- [ ] is_validator is only used in the bond function. Shall we use it elsewhere? unbond?
- [ ] we should think about the logic of slashing and the interplay between withdraw and new_evidence

### DIFF
--- a/2022/Q2/artifacts/pos-model/PoS-model.md
+++ b/2022/Q2/artifacts/pos-model/PoS-model.md
@@ -34,7 +34,7 @@ type Slash struct {
   epoch Epoch
   validator Addr
   block_height int //not used
-  slash_type
+  slash_type {duplicate_vote, ligth_client_attack}
 }
 
 type WeightedValidator struct {
--- a/2022/Q2/artifacts/pos-model/PoS-model.md
+++ b/2022/Q2/artifacts/pos-model/PoS-model.md
@@ -79,8 +79,8 @@ total_voting_power[] in Epoch to VotingPower //map from epoch to voting_power
 tx_become_validator(validator_address, consensus_key, staking_reward_address)
 {
   //check that become_validator has not been called before for validator_address
-  var cur_state = read_epoched_field(validators[validator_address].state, cur_epoch+pipeline_length)
-  if (cur_state == ⊥ && validator_address != staking_reward_address) then
+  var state = read_epoched_field(validators[validator_address].state, cur_epoch+pipeline_length, ⊥)
+  if (state == ⊥ && validator_address != staking_reward_address) then
     validators[validator_address].reward_address = staking_reward_address
     //set status to candidate and consensus key at n + pipeline_length
     validators[validator_address].consensus_key[cur_epoch+pipeline_length] = consensus_key
@@ -89,10 +89,14 @@ tx_become_validator(validator_address, consensus_key, staking_reward_address)
 ```
 
 ```go
+/* COMMENT
+  - https://github.com/informalsystems/partnership-heliax/issues/10
+    issue regarding bonds and deactivation
+*/
 tx_deactivate(validator_address)
 {
-  var cur_state = read_epoched_field(validators[validator_address].state, cur_epoch+pipeline_length)
-  if (cur_state == candidate) then
+  var state = read_epoched_field(validators[validator_address].state, cur_epoch+pipeline_length, ⊥)
+  if (state == candidate) then
     //set status to inactive at n + pipeline_length
     validators[validator_address].state[cur_epoch+pipeline_length] = inactive
 }
@@ -101,8 +105,8 @@ tx_deactivate(validator_address)
 ```go
 tx_reactivate(validator_address)
 {
-  var cur_state = read_epoched_field(validators[validator_address].state, cur_epoch+pipeline_length)
-  if (cur_state == inactive) then
+  var state = read_epoched_field(validators[validator_address].state, cur_epoch+pipeline_length, ⊥)
+  if (state == inactive) then
     //set status to candidate at n + pipeline_length
     validators[validator_address].state[cur_epoch+pipeline_length] = candidate
 }
@@ -152,7 +156,7 @@ tx_delegate(validator_address, delegator_address, amount)
 ```go
 tx_undelegate(validator_address, delegator_address, amount)
 {
-  unbond(src_validator_address, delegator_address, amount)
+  unbond(validator_address, delegator_address, amount)
 }
 ```
 
@@ -177,20 +181,20 @@ tx_withdraw_unbonds_delegator(delegator_address)
 
 ```go
 //This function is called by transactions tx_self_bond, tx_delegate and tx_redelegate
-//the only possible values for offset_length are pipeline_length and ubonding_length
+//the only possible values for offset_length are pipeline_length and unbonding_length
 func bond(validator_address, delegator_address, amount, offset_length)
 {
   if is_validator(validator_address, cur_epoch+offset_length) then
     //add amount bond to delta at n+offset_length
     bonds[delegator_address][validator_address].deltas[cur_epoch+offset_length] += amount
-    //debit amount form delegator account and credit it to the PoS account
+    //debit amount from delegator account and credit it to the PoS account
     balances[delegator_address] -= amount
     balances[pos] += amount
     //compute new total_deltas and write it at n+offset_length
-    var total = total_deltas_at(validators[validator_address].total_deltas, cur_epoch+offset_length)
+    var total = read_epoched_field(validators[validator_address].total_deltas, cur_epoch+offset_length, 0)
     validators[validator_address].total_deltas[cur_epoch+offset_length] = total + amount
     //update validator's voting_power, total_voting_power and validator_sets at n+offset_length
-    power_before = total_deltas_at(validators[validator_address].voting_power, cur_epoch+offset_length)
+    power_before = read_epoched_field(validators[validator_address].voting_power, cur_epoch+offset_length, 0)
     power_after = update_voting_power(validator_address, cur_epoch+offset_length)
     update_total_voting_power(cur_epoch+offset_length)
     update_validator_sets(validator_address, cur_epoch+offset_length, power_before, power_after)
@@ -199,11 +203,11 @@ func bond(validator_address, delegator_address, amount, offset_length)
 
 ```go
 /* COMMENT
-  two issues with the current form:
+  two issues:
   - https://github.com/informalsystems/partnership-heliax/issues/6 (intended)
     delbonds and epoch_counter are considered from the unbonding_length offset.
     This could lead to scenarios in which one starts unbonding before a bond is materialized.
-    Fursthermore, there cannot be positive bonds beyond cur_epoch + pipeline_length, so it does not
+    Furthermore, there cannot be positive bonds beyond cur_epoch + pipeline_length, so it does not
     make sense for delbonds.
   - https://github.com/informalsystems/partnership-heliax/issues/7 (unresolved)
     there is a problem with unbonding, slashing and total_deltas becoming negative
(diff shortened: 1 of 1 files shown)


2022/Q2/artifacts/pos-model/PoS-model.md
```diff
@@ -69,35 +78,21 @@ total_voting_power[] in Epoch to VotingPower //map from epoch to voting_power
...
-  //COMMENT: I do not see in the code where they check that the consensus key is not an old one (this is in the docs).
-  //COMMENT: todo: check there are no consensus_key changes schedule between cur_epoch and the offset. If this necessary though? should
-  //not be enough with the state checks?
-  //https://github.com/anoma/anoma/blob/master/proof_of_stake/src/validation.rs#L515
-  var pre_state = read_epoched_field(validators[validator_address].state, cur_epoch)
-  if (pre_state == ⊥ && validator_address != staking_reward_address) then
-    //reward_address is not in the docs/spec validator struct
+  //check that become_validator has not been called before for validator_address
+  var cur_state = read_epoched_field(validators[validator_address].state, cur_epoch+pipeline_length)
+  if (cur_state == ⊥ && validator_address != staking_reward_address) then
```
sergio-mena: I like the new condition in that it just makes the needed checks that an Anoma transaction would need to do in order not to crash... but it something's wrong with the input parameters, it just becomes a noop.
This will probably help us because now we can add a VP to re-check and provide errors.
This way our model is closer to the actual Anoma exec model

2022/Q2/artifacts/pos-model/PoS-model.md
```diff
@@ -69,35 +78,21 @@ total_voting_power[] in Epoch to VotingPower //map from epoch to voting_power
...
-  //COMMENT: In the code they check not only at cur_epoch but at all epochs between cur_epoch and the offset. Double-check
-  //COMMENT: I do not see in the code where they check that the consensus key is not an old one (this is in the docs).
-  //COMMENT: todo: check there are no consensus_key changes schedule between cur_epoch and the offset. If this necessary though? should
-  //not be enough with the state checks?
-  //https://github.com/anoma/anoma/blob/master/proof_of_stake/src/validation.rs#L515
-  var pre_state = read_epoched_field(validators[validator_address].state, cur_epoch)
-  if (pre_state == ⊥ && validator_address != staking_reward_address) then
-    //reward_address is not in the docs/spec validator struct
+  //check that become_validator has not been called before for validator_address
+  var cur_state = read_epoched_field(validators[validator_address].state, cur_epoch+pipeline_length)
```
sergio-mena: Very minor. I would rename `cur_state`, since it's the state at `cur_epoch+pipeline_length` and not at `cur_epoch`, so I find it a bit misleading.
Maybe just `state`? or `pipeline_state`? or something along those lines?

2022/Q2/artifacts/pos-model/PoS-model.md
```diff
@@ -106,13 +101,8 @@ tx_deactivate(validator_address)
...
 tx_reactivate(validator_address)
 {
-  //VP:https://github.com/anoma/anoma/blob/master/proof_of_stake/src/validation.rs#L459
-  //https://github.com/anoma/anoma/blob/master/proof_of_stake/src/validation.rs#L471
-  var pre_state = read_epoched_field(validators[validator_address].state, cur_epoch)
-  var pre_state_offset = read_epoched_field(validators[validator_address].state, cur_epoch+pipeline_length)
-  if (pre_state == inactive && pre_state_offset in {pending, inactive}) then
-    //set status to pending inmediately
-    validators[validator_address].state[cur_epoch] = pending
+  var cur_state = read_epoched_field(validators[validator_address].state, cur_epoch+pipeline_length)
```
sergio-mena: I really like how the three actions related to validator state have been simplified, without losing power of expression

2022/Q2/artifacts/pos-model/PoS-model.md
```diff
@@ -134,9 +124,11 @@ tx_unbond(validator_address, amount)
...
 tx_withdraw_unbonds_validator(validator_address)
 {
-  //COMMNET: in the docs, the system panics if no self-unbonds. Why? In our view, if a user
-  //attemps to withdraw but has no self-unbounds, the the transaction is a noop. Is there something
-  //wrong about assuming that?
+  /* COMMENT
+  in the docs, the system panics if no self-unbonds. Why? In our view, if a user
+  attemps to withdraw but has no self-unbounds, the the transaction is a noop. Is there something
+  wrong about assuming that? same for tx_withdraw_unbonds_delegator
+  */
```
sergio-mena, line 135: In line 155 below: s/src_validator_address/validator_address/
(thanks Github for the amazing feature set -_-)

2022/Q2/artifacts/pos-model/PoS-model.md
```diff
@@ -191,7 +180,7 @@ tx_withdraw_unbonds_delegator(delegator_address)
 //the only possible values for offset_length are pipeline_length and ubonding_length
```
sergio-mena:
```suggestion
//the only possible values for offset_length are pipeline_length and unbonding_length
```

2022/Q2/artifacts/pos-model/PoS-model.md
```diff
@@ -191,7 +180,7 @@ tx_withdraw_unbonds_delegator(delegator_address)
...
 func bond(validator_address, delegator_address, amount, offset_length)
 {
-  if is_validator(validator_address) then
+  if is_validator(validator_address, cur_epoch+offset_length) then
```
sergio-mena, line 187: I'm confused here, should the target epoch be `cur_epoch+pipeline_length` ?

2022/Q2/artifacts/pos-model/PoS-model.md
```diff
@@ -209,21 +198,26 @@ func bond(validator_address, delegator_address, amount, offset_length)
 ```
 
 ```go
+/* COMMENT
```
sergio-mena, line 205: Because of line 193 above, where we apply `total_deltas_at` to the voting power, consider renaming `total_deltas_at` to something more general

2022/Q2/artifacts/pos-model/PoS-model.md
```diff
@@ -209,21 +198,26 @@ func bond(validator_address, delegator_address, amount, offset_length)
...
+  two issues with the current form:
+  - https://github.com/informalsystems/partnership-heliax/issues/6 (intended)
+    delbonds and epoch_counter are considered from the unbonding_length offset.
+    This could lead to scenarios in which one starts unbonding before a bond is materialized.
+    Fursthermore, there cannot be positive bonds beyond cur_epoch + pipeline_length, so it does not
```
sergio-mena:
```suggestion
    Furthermore, there cannot be positive bonds beyond cur_epoch + pipeline_length, so it does not
```

2022/Q2/artifacts/pos-model/PoS-model.md
```diff
@@ -209,21 +198,26 @@ func bond(validator_address, delegator_address, amount, offset_length)
...
+    make sense for delbonds.
+  - https://github.com/informalsystems/partnership-heliax/issues/7 (unresolved)
+    there is a problem with unbonding, slashing and total_deltas becoming negative
+*/
```
sergio-mena, line 214: Should we turn the "panic" in line 218 into noop, as we are now doing above?

2022/Q2/artifacts/pos-model/PoS-model.md
```diff
@@ -294,15 +276,13 @@ func new_evidence(evidence){
 ```
 
 ```go
-//COMMENT: shall we do updates to this state to happen once at the end of an epoch? This has been discussed.
-//Tomas agreed is interesting, but they are not doing it rigth know.
-//COMMENT: still unclear from the docs. Why are unbonds substracted? When creating unbonds, we already decrement bonds,
-//so we should not substract them here again, unless I am missing something. This is now resolved: there was a typo in the spec.
+/* COMMENT
+shall we do updates to this state to happen once at the end of an epoch? This has been discussed.
+Tomas agreed is interesting, but they are not doing it rigth know.
```
sergio-mena: You may want to add to this comment that doing it at the end may save gas for repetitive transactions in the same epoch

sergio-mena (approved):
Thanks Manu for updating the document. No major issues in my review

2022/Q2/artifacts/pos-model/PoS-model.md
angbrav: Sure

2022/Q2/artifacts/pos-model/PoS-model.md
sergio-mena: Thanks for the off-line explanation. This is now clear to me

2022/Q2/artifacts/pos-model/PoS-model.md
angbrav: yes, you're right. Solved

2022/Q2/artifacts/pos-model/PoS-model.md
angbrav: I am not sure about it. The problem with doing it at the end is that one has to iterate over all validators. While this will be amortized if most of the validators execute PoS transactions in a given epoch, it will be expensive if in practice only a few do. I was actually thinking of removing the comment completely, given that Chris said it was important for them to avoid those iterations. What do you think?

angbrav:
> Thanks Manu for updating the document. No major issues in my review

2022/Q2/artifacts/pos-model/PoS-model.md
sergio-mena: Hmmm. I think I agree. Hadn't thought about it from that perspective

2022/Q2/artifacts/pos-model/PoS-model.md
```diff
@@ -69,35 +78,21 @@ total_voting_power[] in Epoch to VotingPower //map from epoch to voting_power
...
-  if (pre_state == ⊥ && validator_address != staking_reward_address) then
-    //reward_address is not in the docs/spec validator struct
+  //check that become_validator has not been called before for validator_address
+  var cur_state = read_epoched_field(validators[validator_address].state, cur_epoch+pipeline_length)
+  if (cur_state == ⊥ && validator_address != staking_reward_address) then
     validators[validator_address].reward_address = staking_reward_address
-    //set status to pending inmediately
-    validators[validator_address].state[cur_epoch] = pending
     //set status to candidate and consensus key at n + pipeline_length
     validators[validator_address].consensus_key[cur_epoch+pipeline_length] = consensus_key
```
jmalicevic, line 86: So now we do not set the status to anything between `cur_epoch` and `cur_epoch+pipeline_length`?

2022/Q2/artifacts/pos-model/PoS-model.md
jmalicevic, line 187: How will you know that the validator is a validator at` cur_epoch + offset_length` if we only set it to candidate status in a previous `prev_epoch` (`prev_epoch < cur_epoch`) and make it a candidate at `prev_epoch + pipeline_legnth` ?

2022/Q2/artifacts/pos-model/PoS-model.md
sergio-mena: That's right. This is the conclusion we got at in the meeting with Tomas this week.
Rationale is: there's nothing we can do to change the _actual_ state of the current epoch, so changing it in the store introduces a mismatch (and confusion).
And (my own thought here): setting the current epoch to _pending_ does not add any information. If you are _inactive_ (or _bottom_) in the current epoch and you set `cur_epoch+pipeline_length` to _candidate_, and observer of the history will conclude you are efectively in "pending" state in the current epoch.

Side effect: conditions got _so much simpler_ in all three transaction types! (`become_validator`, `deactivate`, `reactivate`)

2022/Q2/artifacts/pos-model/PoS-model.md
sergio-mena: Not sure I totally understood but this may help: function `is_validator` uses function `read_epoched_field`

sergio-mena: Actually, your comment made my reflect about the following.
When we run `tx_deactivate`, we don't look up all the possible `bond` operations that are still ongoing (i.e., no unbond done yet).
So, all those bonds are somehow dangling. The model works because, `tx_undelegate` and `tx_unbond` don't check the status, so the delegator can use them to get the tokens back.
However, what does it mean to have an "ongoing" bonding on a validator that was deactivated long ago? And, when the delegator finally executes `tx_undelegate`, shall it still wait for the unbonding period?

angbrav: Good point. I have opened an issue about this: https://github.com/informalsystems/partnership-heliax/issues/10
 
I think this should not be an issue though. I guess it is up to the users to realize that the validator is not active anymore and claim their tokens back. Regarding whether to still wait for the unbonding period or not, I guess that the rule should be to wait at least unbonding_length from the epoch the validator became inactive. So there should be ways to optimize the model, but it is "safe" in its current shape.

Let's move this discussion to the issue, so we have it all in the same place :)

PR #12, opened 2022-05-17
First TLA+ spec for Anoma's PoS Model

angbrav:
This PR includes a very simplified TLA+ spec of Anoma's PoS Model. 

It DOES model:
- A single validator and any number of users
- The functions delegate, unbond and withdraw
- Epochs
- Things happen immediately

It DOES NOT model:
- Slashing, Jailing, Rewards (still under design)
- Delta's data structures
- Dealing with offsets: pipeline_lenght and unbonding_length
- Time -- epochs advanced based on the number of executed txs.

It includes an instantiation (MC_Staking.tla) that has a simple invariant: balances do not go negative.

### DIFF
--- a/2022/Q2/artifacts/pos-model/MC_Staking.tla
+++ b/2022/Q2/artifacts/pos-model/MC_Staking.tla
@@ -46,4 +46,8 @@ INSTANCE Staking
 BalanceAlwaysPositive == 
     \A user \in UserAddrs: balanceOf[user] >= 0
 
+\* takes forever to check this
+UserConstantAmount == 
+    \A user \in UserAddrs: balanceOf[user] + unbonded[user] + delegated[user] = INITIAL_SUPPLY
+
 ===============================================================================
--- a/2022/Q2/artifacts/pos-model/MC_Staking.tla
+++ b/2022/Q2/artifacts/pos-model/MC_Staking.tla
@@ -1,0 +1,49 @@
+--------------------------- MODULE MC_Staking ---------------------------------
+\* an instance for model checking Staking.tla with Apalache
+EXTENDS Sequences, Staking_typedefs
+
+\* Use the set of three addresses.
+UserAddrs == { "user2", "user3", "validator" }
+
+VARIABLES
+    \* Coin balance for every Cosmos account.
+    \*
+    \* @type: BALANCE;
+    balanceOf,
+    \* Balance of unbonded coins that cannot be used during the bonding period.
+    \*
+    \* @type: BALANCE;
+    unbonded,
+    \* Coins that are delegated to Validator.
+    \*
+    \* @type: BALANCE;
+    delegated
+
+\* Variables that model transactions, not the state machine.
+VARIABLES    
+    \* The last executed transaction.
+    \*
+    \* @type: TX;
+    lastTx,
+    \* A serial number to assign unique ids to transactions
+    \* @type: Int;
+    nextTxId,
+    \* Counts the transaactions executed within an epoch
+    \* @type: Int;
+    txCounter,
+    \* A serial number used to identify epochs
+    \* @type: Int;
+    epoch,
+    \* Whether at least one transaction has failed
+    \* @type: Bool;
+    failed
+
+\* instantiate the spec
+INSTANCE Staking
+
+\* invariants to check and break the system
+
+BalanceAlwaysPositive == 
+    \A user \in UserAddrs: balanceOf[user] >= 0
+
+===============================================================================
--- a/2022/Q2/artifacts/pos-model/Staking.tla
+++ b/2022/Q2/artifacts/pos-model/Staking.tla
@@ -32,7 +32,7 @@ VARIABLES
     \* @type: BALANCE;
     delegated
 
-\* Variables that model transactions, not the state machine.
+\* Variables that model transactions, epochs and offsets, not the state machine.
 VARIABLES    
     \* The last executed transaction.
     \*
@@ -41,7 +41,7 @@ VARIABLES
     \* A serial number to assign unique ids to transactions
     \* @type: Int;
     nextTxId,
-    \* Counts the transaactions executed within an epoch
+    \* Counts the transactions executed within an epoch
     \* @type: Int;
     txCounter,
     \* A serial number used to identify epochs
@@ -68,7 +68,6 @@ Validator == "validator"
 
 \* Initialize the balances
 Init ==
-    \* user{1,2,3} have 1M Cosmos coins, the validator has 1M - stake
     /\ balanceOf = [ a \in UserAddrs |->
         IF a /= "validator"
         THEN INITIAL_SUPPLY
@@ -133,7 +132,7 @@ Unbond(sender, amount) ==
 Withdraw(sender) ==
     LET fail ==
         \/ sender = Validator
-        \/ unbonded[sender] > 0
+        \/ unbonded[sender] <= 0
     IN
     /\ lastTx' = [ id |-> nextTxId, tag |-> "withdraw", fail |-> fail,
                    sender |-> sender, toAddr |-> Validator, value |-> unbonded[sender] ]
--- a/2022/Q2/artifacts/pos-model/Staking.tla
+++ b/2022/Q2/artifacts/pos-model/Staking.tla
@@ -1,0 +1,179 @@
+------------------------- MODULE Staking ---------------------------------
+(*
+ * Modeling the Anoma's staking module.
+ * This is a very minimal spec that includes delegate, unbond and withdraw
+ *
+ * Simplifications:
+ *   - We assume only one validator to be delegated to.
+ *   - Rewards are not specified.
+ *   - Jailing is not specified.
+ *   - Slashing and evidence handling is not specified.
+ * 
+ * Manuel Bravo, 2022
+ *)
+EXTENDS Integers, Apalache, Staking_typedefs
+
+CONSTANTS
+    \* Set of all addresses on Cosmos.
+    \* @type: Set(ADDR);
+    UserAddrs
+
+VARIABLES
+    \* Coin balance for every Cosmos account.
+    \*
+    \* @type: BALANCE;
+    balanceOf,
+    \* Balance of unbonded coins that cannot be used during the bonding period.
+    \*
+    \* @type: BALANCE;
+    unbonded,
+    \* Coins that are delegated to Validator.
+    \*
+    \* @type: BALANCE;
+    delegated
+
+\* Variables that model transactions, not the state machine.
+VARIABLES    
+    \* The last executed transaction.
+    \*
+    \* @type: TX;
+    lastTx,
+    \* A serial number to assign unique ids to transactions
+    \* @type: Int;
+    nextTxId,
+    \* Counts the transaactions executed within an epoch
+    \* @type: Int;
+    txCounter,
+    \* A serial number used to identify epochs
+    \* @type: Int;
+    epoch,
+    \* Whether at least one transaction has failed
+    \* @type: Bool;
+    failed
+
+\* the maximum value in Cosmos
+MAX_UINT == 2^(256 - 60) - 1
+
+\* 1 billion coins in the initial supply
+INITIAL_SUPPLY == 10^(9+18)
+
+\* the number of coins the validator has staked
+INIT_VALIDATOR_STAKE == 1000000000000000000000
+
+\* tx per epoch
+TXS_PER_EPOCH == 10
+
+\* the validator account
+Validator == "validator"
+
+\* Initialize the balances
+Init ==
+    \* user{1,2,3} have 1M Cosmos coins, the validator has 1M - stake
+    /\ balanceOf = [ a \in UserAddrs |->
+        IF a /= "validator"
+        THEN INITIAL_SUPPLY
+        ELSE INITIAL_SUPPLY - INIT_VALIDATOR_STAKE
+       ]
+    /\ unbonded = [ a \in UserAddrs |-> 0 ]
+    /\ delegated = [
+        a \in UserAddrs |-> IF a /= "validator"
+        THEN 0
+        ELSE INIT_VALIDATOR_STAKE
+       ]
+    /\ nextTxId = 0
+    /\ epoch = 1
+    /\ txCounter = 0
+    /\ lastTx = [ id |-> 0, tag |-> "None", fail |-> FALSE ]
+    /\ failed = FALSE
+
+
+\* delegate `amount` coins to Validator
+Delegate(sender, amount) ==
+    LET fail ==
+        \/ amount < 0
+        \/ amount > balanceOf[sender]
+    IN
+    /\ lastTx' = [ id |-> nextTxId, tag |-> "delegate", fail |-> fail,
+                   sender |-> sender, toAddr |-> Validator, value |-> amount ]
+    /\ nextTxId' = nextTxId + 1
+    /\ failed' = (fail \/ failed)
+    /\  IF fail
+        THEN
+          UNCHANGED <<balanceOf, unbonded, delegated>>
+        ELSE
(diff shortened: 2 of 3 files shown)


2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -0,0 +1,179 @@
...
+    epoch,
+    \* Whether at least one transaction has failed
+    \* @type: Bool;
+    failed
+
+\* the maximum value in Cosmos
+MAX_UINT == 2^(256 - 60) - 1
+
+\* 1 billion coins in the initial supply
+INITIAL_SUPPLY == 10^(9+18)
```
sergio-mena, line 58: I don't get the "+18"

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -0,0 +1,179 @@
...
+
+\* the number of coins the validator has staked
+INIT_VALIDATOR_STAKE == 1000000000000000000000
```
sergio-mena, line 61:
```suggestion
INIT_VALIDATOR_STAKE == 10^21
```

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -0,0 +1,179 @@
+------------------------- MODULE Staking ---------------------------------
+(*
+ * Modeling the Anoma's staking module.
+ * This is a very minimal spec that includes delegate, unbond and withdraw
+ *
+ * Simplifications:
+ *   - We assume only one validator to be delegated to.
+ *   - Rewards are not specified.
+ *   - Jailing is not specified.
+ *   - Slashing and evidence handling is not specified.
```
sergio-mena, line 10: I'd add a line saying time is not specified either. In their model, epochs change based on a combination of  transaction number (modeled below) and time

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -0,0 +1,179 @@
...
+INITIAL_SUPPLY == 10^(9+18)
+
+\* the number of coins the validator has staked
+INIT_VALIDATOR_STAKE == 1000000000000000000000
+
+\* tx per epoch
+TXS_PER_EPOCH == 10
+
+\* the validator account
+Validator == "validator"
```
sergio-mena, line 67: Just for my own instruction: should this rather be in MC_Staking.tla?

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -0,0 +1,179 @@
...
+    \*
+    \* @type: BALANCE;
+    delegated
+
+\* Variables that model transactions, not the state machine.
+VARIABLES    
+    \* The last executed transaction.
+    \*
+    \* @type: TX;
+    lastTx,
```
sergio-mena, line 40: Also for my own instruction: do we need to repeat this section here and in MC_Staking.tla ?

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -0,0 +1,179 @@
...
+        IF a /= "validator"
+        THEN INITIAL_SUPPLY
+        ELSE INITIAL_SUPPLY - INIT_VALIDATOR_STAKE
+       ]
+    /\ unbonded = [ a \in UserAddrs |-> 0 ]
+    /\ delegated = [
+        a \in UserAddrs |-> IF a /= "validator"
+        THEN 0
+        ELSE INIT_VALIDATOR_STAKE
+       ]
```
sergio-mena, line 81: Trying to understand...
This will be nuked once we model validator actions, won't it? (except line 77)

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -0,0 +1,179 @@
...
+    /\ nextTxId = 0
+    /\ epoch = 1
+    /\ txCounter = 0
+    /\ lastTx = [ id |-> 0, tag |-> "None", fail |-> FALSE ]
```
sergio-mena, line 85: What happens to `sender` `toAddr` and `value` here, since you don't mention it?

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -0,0 +1,179 @@
...
+                [ balanceOf EXCEPT ![sender] = @ - amount]
+          /\ delegated' = [ delegated EXCEPT ![sender] = @ + amount ]
+          /\ UNCHANGED unbonded
+
+
+\* unbond `amount` coins from Validator
+Unbond(sender, amount) ==
+    LET fail ==
+        \/ amount < 0
+        \/ sender = Validator
```
sergio-mena, line 115: Again, this will be nuked as we extend the model, won't it?

2022/Q2/artifacts/pos-model/MC_Staking.tla
```diff
@@ -0,0 +1,49 @@
...
+    failed
+
+\* instantiate the spec
+INSTANCE Staking
+
+\* invariants to check and break the system
+
+BalanceAlwaysPositive == 
+    \A user \in UserAddrs: balanceOf[user] >= 0
+
```
sergio-mena, line 48: You could also check that \A user balanceOf[user] + delegated[user] = INITIAL_SUPPLY

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -0,0 +1,179 @@
...
+          /\ balanceOf' =
+                [ balanceOf EXCEPT ![sender] = @ - amount]
+          /\ delegated' = [ delegated EXCEPT ![sender] = @ + amount ]
+          /\ UNCHANGED unbonded
+
+
+\* unbond `amount` coins from Validator
+Unbond(sender, amount) ==
+    LET fail ==
+        \/ amount < 0
```
sergio-mena, line 114:
```suggestion
        \/ amount <= 0
```
Do we want to allow useless transactions?

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -0,0 +1,179 @@
...
+    /\ epoch = 1
+    /\ txCounter = 0
+    /\ lastTx = [ id |-> 0, tag |-> "None", fail |-> FALSE ]
+    /\ failed = FALSE
+
+
+\* delegate `amount` coins to Validator
+Delegate(sender, amount) ==
+    LET fail ==
+        \/ amount < 0
```
sergio-mena, line 92:
```suggestion
        \/ amount <= 0
```
Do we want to allow useless transactions?

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -0,0 +1,179 @@
...
+          \* transaction succeeds
+          /\ unbonded' = [ unbonded EXCEPT ![sender] = @ + amount ]
+          /\ delegated' = [ delegated EXCEPT ![sender] = @ - amount ]
+          /\ UNCHANGED  balanceOf
+
+\* withdraw unbonded coins
+Withdraw(sender) ==
+    LET fail ==
+        \/ sender = Validator
+        \/ unbonded[sender] > 0
```
sergio-mena: Are you sure this is what we want here?
Shouldn't the condition be "less than" ?

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -0,0 +1,179 @@
...
+        THEN
+          UNCHANGED <<balanceOf, unbonded, delegated>>
+        ELSE
+          \* transaction succeeds
+          /\ balanceOf' = [ balanceOf EXCEPT ![sender] = @ + unbonded[sender] ]
+          /\ unbonded' = [ unbonded EXCEPT ![sender] = 0 ]
+          /\ UNCHANGED  delegated
+
+AdvanceEpoch ==
+    /\  IF txCounter = TXS_PER_EPOCH
```
sergio-mena, line 151: (just learning...) Is the leading "/\\" necessary here?

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -0,0 +1,179 @@
...
+        THEN
+          \* move to the next epoch
+          /\ epoch' = epoch + 1
+          /\ txCounter' = 0
+        ELSE
+          \* do not advance epoch
+          /\ txCounter' = txCounter + 1
+          /\ UNCHANGED  epoch
```
sergio-mena, line 159: I'm not sure, but I think this doesn't belong here. This will make the counter advance independenly of the actual transactions processed.
I think this should rather be embedded in each of the transactions

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -0,0 +1,179 @@
...
+
+\* The transition relation, which chooses one of the actions
+Next ==
+    \/ \E sender \in UserAddrs:
```
sergio-mena, line 163: I don't get the leading "\\/"

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -0,0 +1,179 @@
...
+       \E amount \in Int:
+        /\ amount <= MAX_UINT
+        /\ \/ Delegate(sender, amount)
+           \/ Unbond(sender, amount)
+           \/ Withdraw(sender)
+        /\ AdvanceEpoch
```
sergio-mena, line 169: (related to the question above) How do you ensure that "AdvanceEpoch" is interleaved one-to-one with one transaction?

2022/Q2/artifacts/pos-model/Staking.tla
konnov: That's the artifact of TLA+. As instantiation is done via substitution, we always have to provide the state variables (or their substitutes) when instantiating a module. Another approach would be to define another module that contains variable definitions and extend it in both `Staking.tla` and `MC_Staking.tla`, but that makes it harder to read both specs.

2022/Q2/artifacts/pos-model/Staking.tla
konnov: The question is whether the implementation allows it. As far as I remember, you can make a transfer of zero coins.

2022/Q2/artifacts/pos-model/Staking.tla
konnov: This can be different in Anoma. This was derived from the Evmos spec, where they have 18 digits after the decimal point.

2022/Q2/artifacts/pos-model/Staking.tla
konnov: You can introduce the set of validators in `CONSTANT` and instantiate it in `MC_Staking.tla`.

2022/Q2/artifacts/pos-model/Staking.tla
konnov: Since we have `/\` in lines 166-170, `AdvanceEpoch` is effectively done in lock step with one of the actions: `Delegate`, `Unbond`, or `Withdraw`.

konnov (commented):
Looks like a good start to me. It would be also useful to define the properties that you want to get checked.

2022/Q2/artifacts/pos-model/Staking.tla
sergio-mena: Thanks Igor. That's clearer now. Cool stuff

2022/Q2/artifacts/pos-model/Staking.tla
sergio-mena: I see. This is probably an example of cases that the model can bring up to Heliax folks so they can decide what they want to do (both spec and implementation-wise)

2022/Q2/artifacts/pos-model/Staking.tla
sergio-mena: Of course! I get it now

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -0,0 +1,179 @@
...
+          \* transaction succeeds
+          \* update the balance of 'sender'
+          /\ balanceOf' =
+                [ balanceOf EXCEPT ![sender] = @ - amount]
+          /\ delegated' = [ delegated EXCEPT ![sender] = @ + amount ]
+          /\ UNCHANGED unbonded
+
+
+\* unbond `amount` coins from Validator
+Unbond(sender, amount) ==
```
jmalicevic, line 112: the sender here is the validator in question?

2022/Q2/artifacts/pos-model/Staking.tla
jmalicevic: Why? I think in the model a validator could also be staking?

2022/Q2/artifacts/pos-model/Staking.tla
jmalicevic: I agree, we cannot withdraw if we have nothing in the unbonded amount?

2022/Q2/artifacts/pos-model/Staking.tla
angbrav: There is a single validator in the current model, so the validator in question does not need to be an argument of the function. The sender is the user that wants to unbond its token. We should discuss whether it makes sense to model multiple validators or not. I am not sure it will add much to the model.

2022/Q2/artifacts/pos-model/Staking.tla
angbrav: Well, not sure if completely, but it will change for sure.

2022/Q2/artifacts/pos-model/Staking.tla
angbrav: you mean the fields of lastTx, right? I am not sure, but I guess they get a default value.

2022/Q2/artifacts/pos-model/Staking.tla
angbrav: > I think in the model a validator could also be staking?

This is prevented in the current model. Again, I think the goal is to find a simplest model that allows us to verify/check the invariants. I am not sure whether self-bond will be important or not.

2022/Q2/artifacts/pos-model/Staking.tla
angbrav: Fixed, thanks!

2022/Q2/artifacts/pos-model/Staking.tla
angbrav: I am not sure I am following. The way `Next` is written, `AdvanceEpoch` is called after a transaction is executed. Therefore, it does not advance independently...

2022/Q2/artifacts/pos-model/Staking.tla
sergio-mena: We'd need to test rebond... so probably need it

2022/Q2/artifacts/pos-model/Staking.tla
sergio-mena: I see... but "id", "tag", and "fail" are also taking default values, aren't they?

2022/Q2/artifacts/pos-model/Staking.tla
sergio-mena: Yes, sorry, this comment was connected to my other comment on `AdvanceEpoch` advancing in lock-step with the different actions. Then, I forgot to come back and delete this one.
Sorry for the noise.

angbrav:
Can I merge this?

jmalicevic:
Yes, all good on my side

sergio-mena:
Go for it!

PR #14, opened 2022-05-25
Add epoched variables

angbrav:
This PR adds epoched variables: `delegated` and `unbonded`. It also add a couple of invariants to generate counter-examples to the model:
- one that generates a trace of 10 transactions
- one that generated a trace that includes at least one `withdraw` transaction

### DIFF
--- a/2022/Q2/artifacts/pos-model/MC_Staking.tla
+++ b/2022/Q2/artifacts/pos-model/MC_Staking.tla
@@ -37,7 +37,7 @@ VARIABLES
     txCounter,
     \* A serial number used to identify epochs
     \* @type: Int;
-    epoch,  
+    epoch,
     \* Whether at least one transaction has failed
     \* @type: Bool;
     failed
--- a/2022/Q2/artifacts/pos-model/MC_Staking.tla
+++ b/2022/Q2/artifacts/pos-model/MC_Staking.tla
@@ -46,4 +46,8 @@ INSTANCE Staking
 BalanceAlwaysPositive == 
     \A user \in UserAddrs: balanceOf[user] >= 0
 
+\* takes forever to check this
+UserConstantAmount == 
+    \A user \in UserAddrs: balanceOf[user] + unbonded[user] + delegated[user] = INITIAL_SUPPLY
+
 ===============================================================================
--- a/2022/Q2/artifacts/pos-model/MC_Staking.tla
+++ b/2022/Q2/artifacts/pos-model/MC_Staking.tla
@@ -1,0 +1,49 @@
+--------------------------- MODULE MC_Staking ---------------------------------
+\* an instance for model checking Staking.tla with Apalache
+EXTENDS Sequences, Staking_typedefs
+
+\* Use the set of three addresses.
+UserAddrs == { "user2", "user3", "validator" }
+
+VARIABLES
+    \* Coin balance for every Cosmos account.
+    \*
+    \* @type: BALANCE;
+    balanceOf,
+    \* Balance of unbonded coins that cannot be used during the bonding period.
+    \*
+    \* @type: BALANCE;
+    unbonded,
+    \* Coins that are delegated to Validator.
+    \*
+    \* @type: BALANCE;
+    delegated
+
+\* Variables that model transactions, not the state machine.
+VARIABLES    
+    \* The last executed transaction.
+    \*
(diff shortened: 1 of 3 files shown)


2022/Q2/artifacts/pos-model/MC_Staking.tla
```diff
@@ -28,12 +32,12 @@ VARIABLES
...
     \* @type: Int;
     nextTxId,
-    \* Counts the transaactions executed within an epoch
+    \* Counts the transactions executed within an epoch
     \* @type: Int;
     txCounter,
     \* A serial number used to identify epochs
     \* @type: Int;
-    epoch,
+    epoch,
```
sergio-mena:
```suggestion
    epoch,
```

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -113,60 +120,78 @@ Unbond(sender, amount) ==
...
     LET fail ==
         \/ sender = Validator
-        \/ unbonded[sender] <= 0
+        \/ unbonded[1, sender] <= 0
     IN
     /\ lastTx' = [ id |-> nextTxId, tag |-> "withdraw", fail |-> fail,
-                   sender |-> sender, toAddr |-> Validator, value |-> unbonded[sender] ]
+                   sender |-> sender, toAddr |-> Validator, value |-> unbonded[1, sender] ]
     /\ nextTxId' = nextTxId + 1
+    /\ txCounter' = txCounter + 1
```
sergio-mena: Would it make more sense to add this in `Next` (so we add it only once for all transaction types), in a similar way we were doing `AdvanceEpoch` before?

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -113,60 +120,78 @@ Unbond(sender, amount) ==
...
     /\ failed' = (fail \/ failed)
+    /\ UNCHANGED epoch
```
sergio-mena: Same here, can we move this to `Next`, with and AND in line 195 ?

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -113,60 +120,78 @@ Unbond(sender, amount) ==
...
+          /\ unbonded' = [ unbonded EXCEPT ![UnbondingLength, sender] = @ + amount ]
+          /\ delegated' = [ delegated EXCEPT ![PipelineLength, sender] = @ - amount ]
           /\ UNCHANGED  balanceOf
 
 \* withdraw unbonded coins
 Withdraw(sender) ==
     LET fail ==
         \/ sender = Validator
-        \/ unbonded[sender] <= 0
+        \/ unbonded[1, sender] <= 0
```
sergio-mena, line 138: why 1?

2022/Q2/artifacts/pos-model/MC_Staking.tla
angbrav: This is just about removing the space, right?

2022/Q2/artifacts/pos-model/Staking.tla
angbrav: I see nothing wrong with doing what you suggest and it is probably neater. It is as it is because Igor's spec was doing it that way and I started from there ;) I have added a new operator Common that includes the actions common to all transactions

2022/Q2/artifacts/pos-model/Staking.tla
angbrav: addressed

2022/Q2/artifacts/pos-model/Staking.tla
angbrav: 1 means cur_epoch. We want to withdraw only tokens already unbonded by cur_epoch

2022/Q2/artifacts/pos-model/MC_Staking.tla
sergio-mena: Yes, small detail

2022/Q2/artifacts/pos-model/Staking.tla
sergio-mena: I see, thanks

sergio-mena (approved):

PR #15, opened 2022-06-08
Adds multiple validators, voting power, and more

angbrav:
This PR:
- Allows the possibility of having more than one validator
- Adds the function `bonded` that records the tokens bonded per validator. This is sort of the voting power of each validator.
- Changes the `delegated` function to record more information: to which validator are tokens delegated per user.
- Makes all epoched variables to be of size unbonding_length
- Fix a couple of issues with dealing with epoched variables
- Adds an invariant that checks that the sum of delegated tokens to a given validator is equal to its voting power. This is, with terms and conditions, what Chris suggested.

It takes forever to check the invariant, so we need to figure out how to optimize the model and/or the invariant. The invariant is currently using Apalache's fold operator.

### DIFF
--- a/2022/Q2/artifacts/pos-model/MC_Staking.tla
+++ b/2022/Q2/artifacts/pos-model/MC_Staking.tla
@@ -8,9 +8,9 @@ UserAddrs == { "user2", "user3", "val1", "val2"}
 \* Set of two validators.
 ValidatorAddrs == {"val1", "val2"}
 
-PipelineLength == 6
+PipelineLength == 2
 
-UnbondingLength == 2
+UnbondingLength == 6
 
 VARIABLES
     \* Coin balance for every Cosmos account.
--- a/2022/Q2/artifacts/pos-model/MC_Staking.tla
+++ b/2022/Q2/artifacts/pos-model/MC_Staking.tla
@@ -37,7 +37,7 @@ VARIABLES
     txCounter,
     \* A serial number used to identify epochs
     \* @type: Int;
-    epoch,  
+    epoch,
     \* Whether at least one transaction has failed
     \* @type: Bool;
     failed
--- a/2022/Q2/artifacts/pos-model/MC_Staking.tla
+++ b/2022/Q2/artifacts/pos-model/MC_Staking.tla
@@ -46,4 +46,8 @@ INSTANCE Staking
 BalanceAlwaysPositive == 
     \A user \in UserAddrs: balanceOf[user] >= 0
 
+\* takes forever to check this
+UserConstantAmount == 
+    \A user \in UserAddrs: balanceOf[user] + unbonded[user] + delegated[user] = INITIAL_SUPPLY
+
 ===============================================================================
--- a/2022/Q2/artifacts/pos-model/MC_Staking.tla
+++ b/2022/Q2/artifacts/pos-model/MC_Staking.tla
@@ -1,0 +1,49 @@
+--------------------------- MODULE MC_Staking ---------------------------------
+\* an instance for model checking Staking.tla with Apalache
+EXTENDS Sequences, Staking_typedefs
+
+\* Use the set of three addresses.
+UserAddrs == { "user2", "user3", "validator" }
+
+VARIABLES
+    \* Coin balance for every Cosmos account.
+    \*
+    \* @type: BALANCE;
+    balanceOf,
+    \* Balance of unbonded coins that cannot be used during the bonding period.
+    \*
+    \* @type: BALANCE;
+    unbonded,
+    \* Coins that are delegated to Validator.
+    \*
+    \* @type: BALANCE;
+    delegated
+
+\* Variables that model transactions, not the state machine.
+VARIABLES    
+    \* The last executed transaction.
+    \*
+    \* @type: TX;
+    lastTx,
+    \* A serial number to assign unique ids to transactions
+    \* @type: Int;
+    nextTxId,
+    \* Counts the transaactions executed within an epoch
+    \* @type: Int;
+    txCounter,
+    \* A serial number used to identify epochs
+    \* @type: Int;
+    epoch,
+    \* Whether at least one transaction has failed
+    \* @type: Bool;
+    failed
+
+\* instantiate the spec
+INSTANCE Staking
+
+\* invariants to check and break the system
+
+BalanceAlwaysPositive == 
+    \A user \in UserAddrs: balanceOf[user] >= 0
+
+===============================================================================
--- a/2022/Q2/artifacts/pos-model/MC_Staking.tla
+++ b/2022/Q2/artifacts/pos-model/MC_Staking.tla
@@ -12,6 +12,8 @@ PipelineLength == 2
 
 UnbondingLength == 6
 
+TxsEpoch == 4
+
 VARIABLES
     \* Coin balance for every Cosmos account.
     \*
@@ -19,15 +21,15 @@ VARIABLES
     balanceOf,
     \* Balance of unbonded coins that cannot be used during the bonding period.
     \*
-    \* @type: EPOCHED;
+    \* @type: UNBONDED;
     unbonded,
     \* Coins that are delegated to Validator.
     \*
-    \* @type: DELEGATEDEPOCHED;
+    \* @type: DELEGATED;
     delegated,
     \* Voting power of the validator.
     \*
-    \* @type: EPOCHED;
+    \* @type: BONDED;
     bonded
 
 \* Variables that model transactions, not the state machine.
--- a/2022/Q2/artifacts/pos-model/MC_Staking.tla
+++ b/2022/Q2/artifacts/pos-model/MC_Staking.tla
@@ -2,8 +2,11 @@
 \* an instance for model checking Staking.tla with Apalache
 EXTENDS Sequences, Staking_typedefs
 
-\* Use the set of three addresses.
-UserAddrs == { "user2", "user3", "validator" }
+\* Use the set of four addresses, including two validators.
+UserAddrs == { "user2", "user3", "val1", "val2"}
+
+\* Set of two validators.
+ValidatorAddrs == {"val1", "val2"}
 
 PipelineLength == 6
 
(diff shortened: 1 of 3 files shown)


konnov:
The invariant is checked relatively fast in the `simulate` mode, so your spec is relatively fast. If you want to prove the invariant for all executions of arbitrary lengths, you would have to construct an inductive invariant, see: https://apalache.informal.systems/docs/apalache/running.html#15-checking-an-inductive-invariant

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -93,56 +104,65 @@ Init ==
...
-          /\ delegated' = [ delegated EXCEPT ![PipelineLength, sender] = @ + amount]
+          /\ delegated' = [ n \in 1..UnbondingLength, user \in UserAddrs, val \in ValidatorAddrs |-> 
+                     IF n >= PipelineLength /\ user = sender /\ val = validator
+                     THEN delegated[n, user, val] + amount
+                     ELSE delegated[n, user, val]
+                    ]
+          /\ bonded' = [ n \in 1..UnbondingLength, user \in ValidatorAddrs |-> 
+                     IF n >= PipelineLength /\ user = sender
+                     THEN bonded[n, user] + amount
+                     ELSE bonded[n, user]
```
konnov, line 131: I do not know the logic of this protocol, but is not it a bit suspicious that the balance is reduced and neither `bonded` nor `delegated` get increased when `n < PipelineLength`?

konnov (commented):
Looks like a very clean spec to me!

2022/Q2/artifacts/pos-model/Staking_typedefs.tla
```diff
@@ -22,8 +24,9 @@
   A state of the state machine:
   @typeAlias: STATE = [
     balanceOf: BALANCE,
-    delegated: EPOCHED,
+    delegated: DELEGATEDEPOCHED,
     unbonded: EPOCHED,
```
sergio-mena: Very minor: Rename EPOCHED to BONDEDEPOCHED ?

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -164,11 +184,16 @@ ShiftEpochoedVariables ==
...
+                     THEN delegated[n+1, user, val]
+                     ELSE delegated[n, user, val]
                     ]
+    /\ bonded' = [ n \in 1..UnbondingLength, user \in ValidatorAddrs |-> 
+                  IF n < UnbondingLength
+                  THEN bonded[n+1, user]
+                  ELSE bonded[n, user]
+                 ]
 
 AdvanceEpoch ==
```
sergio-mena, line 194: Minor: variable name "ShiftEpochoedVariables" has a typo

2022/Q2/artifacts/pos-model/MC_Staking.tla
```diff
@@ -2,8 +2,11 @@
...
 
-\* Use the set of three addresses.
-UserAddrs == { "user2", "user3", "validator" }
+\* Use the set of four addresses, including two validators.
+UserAddrs == { "user2", "user3", "val1", "val2"}
+
+\* Set of two validators.
+ValidatorAddrs == {"val1", "val2"}
 
 PipelineLength == 6
```
sergio-mena: Aren't these two lengths swapped?
Shouldn't unbonding length be greater than pipeline length?

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -57,34 +67,35 @@ VARIABLES
...
 
-\* the number of coins the validator has staked
+\* the number of tokens the validator has staked
 INIT_VALIDATOR_STAKE == 1000000000000000000000
 
+\* the amount of voting power per token
+VOTES_PER_TOKEN == 1
+
 \* tx per epoch
 TXS_PER_EPOCH == 10
```
sergio-mena: If we reduce this number to, say, 4... wouldn't it help with the time it takes to check invariants without loss of generality?

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -93,56 +104,65 @@ Init ==
 
 
 \* delegate `amount` coins to Validator
-Delegate(sender, amount) ==
+Delegate(sender, validator, amount) ==
```
sergio-mena, line 108: Do we want to allow (in the model, for efficiency) a validator to delegate to another validator?
More generally, do they (i.e., Heliax) want to allow that in the actual system?

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -14,29 +14,39 @@
...
     \*
     \* @type: BALANCE;
     balanceOf,
-    \* Balance of unbonded coins that cannot be used during the bonding period.
+    \* Balance of unbonded tokens that cannot be used during the bonding period.
     \*
     \* @type: EPOCHED;
     unbonded,
-    \* Coins that are delegated to Validator.
+    \* Token that are delegated to Validator.
```
sergio-mena: Minor: constant "Validator" no longer exists. Please search "Validator" in comments and adapt them if needed

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -93,56 +104,65 @@ Init ==
...
+          /\ bonded' = [ bonded EXCEPT ![UnbondingLength, validator] = @ - amount]
           /\ UNCHANGED  balanceOf
+          
 
-\* withdraw unbonded coins
+\* withdraw unbonded tokens
 Withdraw(sender) ==
     LET fail ==
-        \/ sender = Validator
         \/ unbonded[1, sender] <= 0
```
sergio-mena: [This is just a question of interpretation]
Is this really a failure then the value is 0?
Should it just be a successful withdrawal of 0 tokens? Note that, e.g., in Delegate we allow 0 tokens

2022/Q2/artifacts/pos-model/Staking.tla
```diff
@@ -57,34 +67,35 @@ VARIABLES
     \* @type: Bool;
     failed
 
-\* the maximum value in Cosmos
+\* the maximum value
 MAX_UINT == 2^(256 - 60) - 1
 
-\* 1 billion coins in the initial supply
+\* 1 billion toekns in the initial supply
```
sergio-mena:
```suggestion
\* 1 billion tokens in the initial supply
```

2022/Q2/artifacts/pos-model/Staking.tla
angbrav: That's intended: delegations only get materialized at Pipelinelength. Balances are reduced right away to avoid double-spending.

2022/Q2/artifacts/pos-model/Staking.tla
angbrav: I do not think there is a major difference from the point of view of the model and it is simple this way.

2022/Q2/artifacts/pos-model/Staking.tla
angbrav: It is not really. I think it does not matter, but I'll make it consistent by removing it.

2022/Q2/artifacts/pos-model/MC_Staking.tla
angbrav: You are right. Fixed.

2022/Q2/artifacts/pos-model/Staking_typedefs.tla
angbrav: both unbonded and bonded use EPOCHED type, so I do not think it is a good idea. Maybe we can change the name of DELEGATEDEPOCHED. Any suggestion?

2022/Q2/artifacts/pos-model/Staking.tla
angbrav: The invariants are checked after executing every single transaction, so I do not see how reducing the number of txs per epoch would help the checker

sergio-mena: What I had in mind is that, if we are exploring possible transitions of the model, and most interesting things happen when the epoch changes, reducing the transactions per epoch might help finding counter-examples faster... but I might be missing something

2022/Q2/artifacts/pos-model/Staking.tla
sergio-mena: OK, makes sense. But I think we should ask Heliax about this as a case they should consider what they want... maybe when presenting the model to them

2022/Q2/artifacts/pos-model/Staking_typedefs.tla
sergio-mena: I see... I had missed the detail that `unbonded` was also EPOCHED.

How about:
EPOCHED ==> EPOCHED_BONDS
DELEGATEDEPOCHED ==> EPOCHED_DELEGATIONS

Or, another idea:
EPOCHED ==> ADDR_EPOCH
DELEGATEDEPOCHED ==> ADDR_ADDR_EPOCH

Up to you!

sergio-mena (approved):

2022/Q2/artifacts/pos-model/Staking.tla
angbrav: This will depend on how we write invariants. For now, the invariants use the variables and do not refer to epochs. In any case, I have created a constant, so it is now configurable.

2022/Q2/artifacts/pos-model/Staking_typedefs.tla
angbrav: Fixed it using BONDED, UNBONDED and DELEGATED.

angbrav:
> I would still keep the check "<0" (as in Delegate)... I mean "/ unbonded[1, sender] < 0"

The way I understand those checks at the beginning of the operators is more like preconditions on the operator arguments. The fact that delegations should not go negative should be checked an invariant.

Issue #7, opened 2022-03-30
total_deltas becoming negative


angbrav:
### Potential problem with total_deltas

Following there is an execution that makes `total_deltas` of a validator `val` become negative. The VP code seems to disallow it:
https://github.com/anoma/anoma/blob/master/proof_of_stake/src/validation.rs#L694-L698

- Let `validators[val].total_deltas[e+1-unbonding_length] = a1`
- Assume that `val` decides to unbond a1 at `e+1-unbonding_length`
- We have that `validators[val].total_deltas[e+1] = a1 - a1 = 0`
- Assume now that some `evidence` is found for an action `val` did at epoch `e`
- We compute `slashed_amount = validators[val].total_deltas[e] * slash_rate = a1 * slash_rate`
- Then update `total_deltas` at the offset: `validators[val].total_deltas[e+pipeline_length] = 0 - a1*slash_rate < 0`

tzemanovic:
I think the issue might be even worse than that, because we’re subtracting absolute slashed amounts from the validator total deltas. The validator total deltas are summed in each epoch, so we cannot tell what amount has been bonded and unbonded just from this field alone, so the absolute slashed amounts are effectively slashing the bonds (positive deltas), but not the unbonds (negative deltas), which doesn’t just lead to underflow but also wrong values when slashed bond is partially unbonded and when new bonds are added to be materialized at the same epoch as unbonds (submit a new bond unbonding_len - pipeline_len epochs after unbonding) With the example from the issue, say the slash rate is 50%, but instead of unbonding the full a1 , the validator unbonds just a half. Then what we end up with in the validator total deltas would amount to sum(a1, -a1/2, -a1/2) = 0 , but in fact, the validator should be left with a1/4 (half of the bond that wasn’t unbonded, slashed by 50%).

We're gonna revisit this with new slashing design.

tzemanovic:
[...]

PR #16, opened 2022-07-04
Add cubic slashing to pseudocode model

angbrav:
This PR integrates [cubic slashing](https://specs.anoma.net/master/architecture/economics/proof-of-stake/cubic-slashing.html) to the pseudocode model.

I have created a new file because this PR does not add a new feature but changes the slashing mechanism. This makes it more challenging to review, as the whole file is marked a new by GitHub. Things to review:
- func new_evidence
- func end_of_epoch

Open issues:
- Some epoched variables get updated at different offsets: validator.state and validator_sets

### DIFF
--- a/2022/Q2/artifacts/pos-model/PoS-model-cubic.md
+++ b/2022/Q2/artifacts/pos-model/PoS-model-cubic.md
@@ -1,0 +1,493 @@
+## Assumptions/simplifications
+
+- There is a single token type
+
+## Data types
+
+```go
+type Addr
+type Key
+type Epoch uint
+type VotingPower uint
+
+type Validator struct {
+  consensus_key map<Epoch, Key>
+  state map<Epoch, {inactive, candidate, frozen, jailed}>
+  total_deltas map<Epoch, amount:int>
+  voting_power map<Epoch, VotingPower>
+  reward_address Addr
+}
+
+type Bond struct {
+  validator Addr //not used
+  source Addr //not used
+  deltas map<Epoch, int>
+  }
+
+type Unbond struct {
+  validator Addr //not used
+  source Addr //not used
+  deltas map<(start:Epoch, end:Epoch), int>
+}
+
+type Slash struct {
+  epoch Epoch
+  validator Addr
+  voting_power VotingPower //new in cubic slashing
+  slash_type {duplicate_vote, ligth_client_attack} //not used in cubic slashing
+}
+
+type WeightedValidator struct {
+  validator Addr
+  voting_power VotingPower
+}
+
+type ValidatorSet struct {
+  active orderedset<WeightedValidator>
+  inactive orderedset<WeightedValidator>
+}
+```
+
+## Constants
+
+```go
+pipeline_length uint
+unbonding_length uint
+votes_per_token uint
+duplicate_vote_rate float //not used in cubic slashing
+ligth_client_attack_rate float //not used in cubic slashing
+```
+
+## Variables
+
+```go
+cur_epoch ← 0 in Epoch //current epoch
+
+validators[] in Addr → Validator //map from address to validator
+balances[] in Addr → int //map from address to integer
+bonds[][] in (Addr X Addr) → Bond //map from address to map from address to bond
+unbonds[][] in (Addr X Addr) → Unbond  //map from (address, address) to unbond
+slashes[] in Epoch → 2^Slash //map from address to list of slashes
+
+validator_sets[] in Epoch → ValidatorSet //map from epoch to validator_set
+total_voting_power[] in Epoch to VotingPower //map from epoch to voting_power
+```
+
+## Validator transactions
+
+```go
+tx_become_validator(validator_address, consensus_key, staking_reward_address)
+{
+  //check that become_validator has not been called before for validator_address
+  var state = read_epoched_field(validators[validator_address].state, cur_epoch+pipeline_length, ⊥)
+  if (state == ⊥ && validator_address != staking_reward_address) then
+    validators[validator_address].reward_address = staking_reward_address
+    //set status to candidate and consensus key at n + pipeline_length
+    validators[validator_address].consensus_key[cur_epoch+pipeline_length] = consensus_key
+    validators[validator_address].state[cur_epoch+pipeline_length] = candidate
+}
+```
+
+```go
+/* COMMENT
+  - https://github.com/informalsystems/partnership-heliax/issues/10
+    issue about bonds and deactivation
+    conclusion: leave it as it is for the moment.
+*/
+tx_deactivate(validator_address)
+{
+  var state = read_epoched_field(validators[validator_address].state, cur_epoch+pipeline_length, ⊥)
+  if (state == candidate) then
+    //set status to inactive at n + pipeline_length
+    validators[validator_address].state[cur_epoch+pipeline_length] = inactive
+}
+```
+
+```go
+tx_reactivate(validator_address)
+{
+  var state = read_epoched_field(validators[validator_address].state, cur_epoch+pipeline_length, ⊥)
+  if (state == inactive) then
+    //set status to candidate at n + pipeline_length
+    validators[validator_address].state[cur_epoch+pipeline_length] = candidate
+}
+```
+
+```go
+tx_self_bond(validator_address, amount)
+{
+  bond(validator_address, validator_address, amount, pipeline_length)
+}
+```
+
+```go
+tx_unbond(validator_address, amount)
+{
+  unbond(validator_address, validator_address, amount)
+```
+
+```go
+tx_withdraw_unbonds_validator(validator_address)
+{
+  withdraw(validator_address, validator_address)
+}
+```
+
+```go
+tx_change_consensus_key(validator_address, consensus_key)
+{
+  //set consensus key at n + pipeline_length
+  validators[validator_address].consensus_key[cur_epoch+pipeline_length] = consensus_key
+}
+```
+## Delegator transactions
+
+```go
+tx_delegate(validator_address, delegator_address, amount)
+{
+  bond(validator_address, delegator_address, amount, pipeline_length)
+}
+```
+
+```go
+tx_undelegate(validator_address, delegator_address, amount)
+{
+  unbond(validator_address, delegator_address, amount)
+}
+```
+
+<!-- 
+```go
+tx_redelegate(src_validator_address, dest_validator_address, delegator_address, amount)
+{
+  unbond(src_validator_address, delegator_address, amount)
+  bond(dst_validator_address, delegator_address, amount, unbonding_length)
+}
+```
+-->
+
+```go
+tx_withdraw_unbonds_delegator(delegator_address)
+{
+  forall (validator_address in validators) do
+    withdraw(validator_address, delegator_address)
+}
+```
+
+## PoS functions
+
+
+```go
+//This function is called by transactions tx_self_bond, tx_delegate and tx_redelegate
+//the only possible values for offset_length are pipeline_length and unbonding_length
+func bond(validator_address, delegator_address, amount, offset_length)
+{
+  //QUESTION: What if we remove this and it is up to the user to be smart bonding its tokens
+  if is_validator(validator_address, cur_epoch+offset_length) then
+    //add amount bond to delta at n+offset_length
+    bonds[delegator_address][validator_address].deltas[cur_epoch+offset_length] += amount
+    //debit amount from delegator account and credit it to the PoS account
+    balances[delegator_address] -= amount
+    balances[pos] += amount
+    //compute new total_deltas and write it at n+offset_length
+    var total = read_epoched_field(validators[validator_address].total_deltas, cur_epoch+offset_length, 0)
+    validators[validator_address].total_deltas[cur_epoch+offset_length] = total + amount
+    //update validator's voting_power, total_voting_power and validator_sets at n+offset_length
+    power_before = read_epoched_field(validators[validator_address].voting_power, cur_epoch+offset_length, 0)
+    power_after = update_voting_power(validator_address, cur_epoch+offset_length)
+    update_total_voting_power(cur_epoch+offset_length)
+    update_validator_sets(validator_address, cur_epoch+offset_length, power_before, power_after)
+}
+```
+
+```go
+/* COMMENT
+  two issues:
+  - https://github.com/informalsystems/partnership-heliax/issues/6
+    delbonds and epoch_counter are considered from the unbonding_length offset.
+    This could lead to scenarios in which one starts unbonding before a bond is materialized.
+    Furthermore, there cannot be positive bonds beyond cur_epoch + pipeline_length, so it does not
+    make sense for delbonds.
+    conclusion: intended
+  - https://github.com/informalsystems/partnership-heliax/issues/7
+    there is a problem with unbonding, slashing and total_deltas becoming negative
+    conclusion: it is an actual issue, unresolved
+*/
+//This function is called by transactions tx_unbond, tx_undelegate and tx_redelegate
+func unbond(validator_address, delegator_address, amount)
+{
+  //disallow unbonding if the validator is frozen
+  var state = read_epoched_field(validators[validator_address].state, cur_epoch, ⊥)
+  if (state != frozen) then
+    //compute total bonds from delegator to validator
+    var delbonds = compute_total_from_deltas(bonds[delegator_address][validator_address].deltas, cur_epoch + unbonding_length)
+    //check if there are enough bonds
+    //this serves to check that there are bonds (in the docs) and that these are greater than the amount we are trying to unbond
+    if (delbonds >= amount) then
+      //Decrement bond deltas and create unbonds
+      var remain = amount
+      var epoch_counter = cur_epoch + unbonding_length + 1
+      while remain > 0 do
+        epoch_counter = max{epoch | bonds[delegator_address][validator_address].deltas[epoch] > 0 && epoch < epoch_counter}
+        var bond = bonds[delegator_address][validator_address].deltas[epoch_counter]
+        if bond > remain then var unbond_amount = remain
+        else var unbond_amount = bond
+        unbonds[delegator_address][validator_address].deltas[(epoch_counter,cur_epoch+unbonding_length)] += unbond_amount
+        remain -= unbond_amount
+      bonds[delegator_address][validator_address].deltas[cur_epoch+unbonding_length] -= amount
+      //compute new total_deltas and write it at n+unbonding_length
+      var total = read_epoched_field(validators[validator_address].total_deltas, cur_epoch+unbonding_length, 0)
+      validators[validator_address].total_deltas[cur_epoch+unbonding_length] = total - amount
+      //update validator's voting_power, total_voting_power and validator_sets at n+unbonding_length
+      power_before = read_epoched_field(validators[validator_address].voting_power, cur_epoch+unbonding_length, 0)
+      power_after = update_voting_power(validator_address, cur_epoch+unbonding_length)
+      update_total_voting_power(cur_epoch+unbonding_length)
+      update_validator_sets(validator_address, cur_epoch+unbonding_length, power_before, power_after)
+}
+```
+
+```go
+//This function is called by transactions tx_withdraw_unbonds_validator and tx_withdraw_unbonds_delegator
+func withdraw(validator_address, delegator_address)
+{
+  var delunbonds = {<start,end,amount> | amount = unbonds[delegator_address][validator_address].deltas[(start, end)] > 0 && end <= cur_epoch }
+  //substract any pending slash before withdrawing
+  forall (<start,end,amount> in selfunbonds) do
+    var amount_after_slashing = amount
+    forall (slash in slashes[validator_address] s.t. start <= slash.epoch && slash.epoch <= end)
+      amount_after_slashing += amount*slash.rate
+    balance[delegator_address] += amount_after_slashing
+    balance[pos] -= amount_after_slashing
+    //remove unbond
+    unbonds[delegator_address][validator_address].deltas[(start,end)] = 0
+}
+```
+
+```go
+func new_evidence(evidence)
+{
+  //create slash
+  var vpower = read_epoched_field(validators[evidence.validator].total_deltas, evidence.epoch, 0)
+  var slash = Slash{epoch: evidence.epoch, rate: 0, voting_power: vpower}
+  //enqueue slash (Step 1.1 of cubic slashing)
+  append(enqueued_slashes[evidence.epoch + unbonding_epoch], slash)
+  //freeze validator to prevent delegrators from altering their delegations (Step 1.3 of cubic slashing)
+  validators[validator_address].state[cur_epoch] = frozen
+}
+```
+
+```go
+func update_voting_power(validator_address, epoch)
+{
+  //compute bonds from total_deltas 
+  var bonds = read_epoched_field(validators[validator_address].total_deltas, epoch, 0)
+  //compute the new voting power
+  var power_after = votes_per_token*bonds
+  //update voting power and return it
+  validators[validator_address].voting_power[epoch] = power_after
+  return power_after
+}
+```
+```go
+func update_total_voting_power(epoch)
+{
+  var total = 0
+  forall (validator in validator_sets[epoch].active U validator_sets[epoch].inactive) do
+    total += validator.voting_power
+  total_voting_power[epoch] = total
+}
+```
+```go
+func update_validator_sets(validator_address, epoch, power_before, power_after)
+{
+  var min_active = first(validator_sets[epoch].active)
+  var max_inactive = last(validator_sets[epoch].inactive)
+  if (power_before >= max_inactive.voting_power) then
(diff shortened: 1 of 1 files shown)


2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -0,0 +1,493 @@
...
+## Variables
+
+```go
+cur_epoch ← 0 in Epoch //current epoch
+
+validators[] in Addr → Validator //map from address to validator
+balances[] in Addr → int //map from address to integer
+bonds[][] in (Addr X Addr) → Bond //map from address to map from address to bond
+unbonds[][] in (Addr X Addr) → Unbond  //map from (address, address) to unbond
+slashes[] in Epoch → 2^Slash //map from address to list of slashes
```
tzemanovic:
```suggestion
slashes[] in Epoch → 2^Slash //map from epoch to list of slashes
```

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -0,0 +1,493 @@
...
+
+```go
+type Addr
+type Key
+type Epoch uint
+type VotingPower uint
+
+type Validator struct {
+  consensus_key map<Epoch, Key>
+  state map<Epoch, {inactive, candidate, frozen, jailed}>
```
tzemanovic: I think the `frozen` state may need to be a separate field as it needs to be set as soon as evidence is found

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -0,0 +1,493 @@
...
+  validator Addr //not used
+  source Addr //not used
+  deltas map<(start:Epoch, end:Epoch), int>
+}
+
+type Slash struct {
+  epoch Epoch
+  validator Addr
+  voting_power VotingPower //new in cubic slashing
+  slash_type {duplicate_vote, ligth_client_attack} //not used in cubic slashing
```
sergio-mena: I might have misunderstood, but I think we still have different types of misbehavior

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -0,0 +1,493 @@
...
+  }
+
+type Unbond struct {
+  validator Addr //not used
+  source Addr //not used
+  deltas map<(start:Epoch, end:Epoch), int>
+}
+
+type Slash struct {
+  epoch Epoch
```
sergio-mena, line 37: Do we need this? (it is now the key of the map containing this structure)

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -0,0 +1,493 @@
...
+  - https://github.com/informalsystems/partnership-heliax/issues/7
+    there is a problem with unbonding, slashing and total_deltas becoming negative
+    conclusion: it is an actual issue, unresolved
+*/
+//This function is called by transactions tx_unbond, tx_undelegate and tx_redelegate
+func unbond(validator_address, delegator_address, amount)
+{
+  //disallow unbonding if the validator is frozen
+  var state = read_epoched_field(validators[validator_address].state, cur_epoch, ⊥)
+  if (state != frozen) then
```
sergio-mena: how about jailed?

angbrav: I think this is intended, one should allow delegators to unbond their tokens if the validator gets jalied.

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
angbrav: This is something I want to clarify. In the spec, the infraction type is never used explicitely.

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
angbrav: Thanks, fixed.

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -0,0 +1,493 @@
...
+```
+
+```go
+func new_evidence(evidence)
+{
+  //create slash
+  var vpower = read_epoched_field(validators[evidence.validator].total_deltas, evidence.epoch, 0)
+  var slash = Slash{epoch: evidence.epoch, rate: 0, voting_power: vpower}
+  //enqueue slash (Step 1.1 of cubic slashing)
+  append(enqueued_slashes[evidence.epoch + unbonding_epoch], slash)
```
sergio-mena: I can't find the definition of `enqueued_slashes`

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -0,0 +1,493 @@
...
+```go
+end_of_epoch()
+{
+  set_validators = {val | val = slash.validator && slash in enqueued_slashes[cur_epoch]}
+  forall (validator_address in set_validators) do
+    //jail validator (Step 1.2 of cubic slashing). This also unfreeze the validator (Step 2.5 of cubic slashing)
+    validators[validator_address].state[cur_epoch+1] = jailed
+    remove_from_validator_sets(validator_address, cur_epoch+1)
+    //iterate over all slashes for infractions within (-1,+1) epochs range (Step 2.1 of cubic slashing)
+    var set_slashes = {s | s in enqueued_slashes[epoch] && cur_epoch-1 <= epoch <= cur_epoch+1 && s.validator == validator_address}
```
jmalicevic, line 380: As we are iterating over the previous and next epochs, and we will repeat this step for` cur_epoch + 1`, are we not punishing the validator multiple times for the same evidence. So for evidence from cur_epoch, the validator is punished in epochs` cur_epoch -1`, `cur_epoch `and `cur_epoch + 1`. 


Related to this, at the end of` curr_epoch`, we decrease the voting power (and the total voting power). But then at the end of` curr_epoch + 1 `we will do the same thing taking into account an at that point invalid voting power and decreasing it again.

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -0,0 +1,493 @@
...
+    //calculate the slash rate (Step 2.2 of cubic slashing)
+    var rate = compute_slash_rate(set_slashes)
+    forall (slash in {s | s in enqueued_slashes[cur_epoch] && slash.validator == validator_address}) do
+      //set the slash on the now "finalised" slash in storage (Step 2.3 of cubic slashing)
+      slash.rate = rate
+      append(slashes[cur_epoch], slash)
+      slashed_amount = slash.voting_power * slash.rate
```
sergio-mena: So all misbehavior types have the same _severity_?

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
angbrav: Actually, there are some inconsistencies. Withdrawing is till using the slashes variables indexed by validators. I'll fix this and remove the epoch field if it is not needed. Thanks for pointing it out!

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
sergio-mena: Also, what would happen if `evidence.epoch` is too far in the past? (is it possible?)

angbrav: Fixed.

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
angbrav: Unclear from the spec. Need clarification

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -0,0 +1,493 @@
...
+```
+
+## PoS functions
+
+
+```go
+//This function is called by transactions tx_self_bond, tx_delegate and tx_redelegate
+//the only possible values for offset_length are pipeline_length and unbonding_length
+func bond(validator_address, delegator_address, amount, offset_length)
+{
```
sergio-mena, line 189: Do we need to block this as well for frozen validators?

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
angbrav: That's the behavior now: the validator's state is set to frozen as soon as evidence is found. But I agree this needs more analysis as it may interfere with other behaviors.

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
sergio-mena: Should we _unfreeze_ validators here?
Otherwise, how does a frozen validator leave that state?

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
angbrav: I think I fixed it.

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
sergio-mena: Oh, I see they get jailed. Please ignore my comment

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -0,0 +1,493 @@
...
+```
+
+```go
+end_of_epoch()
+{
+  set_validators = {val | val = slash.validator && slash in enqueued_slashes[cur_epoch]}
+  forall (validator_address in set_validators) do
+    //jail validator (Step 1.2 of cubic slashing). This also unfreeze the validator (Step 2.5 of cubic slashing)
+    validators[validator_address].state[cur_epoch+1] = jailed
+    remove_from_validator_sets(validator_address, cur_epoch+1)
```
jmalicevic: In `bond()`, we allow delegators to bond assets to any validator that is a validator for a current epoch. As the epochs we bond to  can be in the future compared to when the validator is jailed, technically, we allow bonding assets to a validator that is currently jailed. Should we have a condition for bonding to a validator only if this `validotor.state != jailed` in any of the epochs before `bonding_epoch` ?

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -0,0 +1,517 @@
...
+type Unbond struct {
+  validator Addr //not used
+  source Addr //not used
+  deltas map<(start:Epoch, end:Epoch), int>
+}
+
+type Slash struct {
+  epoch Epoch
+  rate float
+  vpower_fraction VotinPower //new in cubic slashing
```
sergio-mena:
```suggestion
  vpower_fraction VotingPower //new in cubic slashing
```

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -264,16 +266,28 @@ func withdraw(validator_address, delegator_address)
...
   var vpower = read_epoched_field(validators[evidence.validator].total_deltas, evidence.epoch, 0)
-  var slash = Slash{epoch: evidence.epoch, rate: 0, voting_power: vpower}
+  var slash = Slash{epoch: evidence.epoch, rate: 0, vpower_fraction: compute_vpower_fraction(evidence.type, vpower)}
   //enqueue slash (Step 1.1 of cubic slashing)
-  append(enqueued_slashes[evidence.epoch + unbonding_epoch], slash)
+  append(enqueued_slashes[evidence.epoch + unbonding_length], slash)
+  //jail validator (Step 1.2 of cubic slashing)
+  validators[validator_address].state[cur_epoch + 1] = jailed
+  remove_from_validator_sets(validator_address, cur_epoch + 1)
   //freeze validator to prevent delegrators from altering their delegations (Step 1.3 of cubic slashing)
```
sergio-mena:
```suggestion
  //freeze validator to prevent delegators from altering their delegations (Step 1.3 of cubic slashing)
```

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -0,0 +1,517 @@
...
+*/
+func new_evidence(evidence)
+{
+  //create slash
+  var vpower = read_epoched_field(validators[evidence.validator].total_deltas, evidence.epoch, 0)
+  var slash = Slash{epoch: evidence.epoch, rate: 0, vpower_fraction: compute_vpower_fraction(evidence.type, vpower)}
+  //enqueue slash (Step 1.1 of cubic slashing)
+  append(enqueued_slashes[evidence.epoch + unbonding_length], slash)
+  //jail validator (Step 1.2 of cubic slashing)
+  validators[validator_address].state[cur_epoch + 1] = jailed
```
sergio-mena: Isn't _jailed_ a different state now? (non-epoched)

sergio-mena:
```suggestion
  validators[validator_address].jailed = true
```

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -0,0 +1,517 @@
...
+```go
+end_of_epoch()
+{
+  set_validators = {val | val = slash.validator && slash in enqueued_slashes[cur_epoch]}
+  forall (validator_address in set_validators) do
+    //iterate over all slashes for infractions within (-1,+1) epochs range (Step 2.1 of cubic slashing)
+    var set_slashes = {s | s in enqueued_slashes[epoch] && cur_epoch-1 <= epoch <= cur_epoch+1 && s.validator == validator_address}
+    //calculate the slash rate (Step 2.2 of cubic slashing)
+    var rate = compute_final_rate(set_slashes)
+    forall (slash in {s | s in enqueued_slashes[cur_epoch] && slash.validator == validator_address}) do
```
sergio-mena: The nested "forall" delimited by indentation (_à la_ python) is a bit confusing. Would it be possible to extract the logic of the inner "forall" into its own function? (or something similar)

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -0,0 +1,517 @@
...
+    else
+      remove(validator_sets[epoch].inactive, validator_address)
+      add(validator_sets[epoch].inactive, <validator_address, power_after>)
+}
+```
+
+```go
+func remove_from_validator_sets(validator_address)
+{
+  var max_inactive = last(validator_sets[epoch].inactive)
```
sergio-mena, line 352: where is `epoch` defined? Is it missing as parameter?

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -0,0 +1,517 @@
...
+      add(validator_sets[epoch].active, <validator_address, power_after>)
+    //if inactive and its change in power does not promote it to active, then update its voting power
+    else
+      remove(validator_sets[epoch].inactive, validator_address)
+      add(validator_sets[epoch].inactive, <validator_address, power_after>)
+}
+```
+
+```go
+func remove_from_validator_sets(validator_address)
```
sergio-mena: Two comments (related to each other):
* Isn't this the job of `update_validator_sets`?
* As we are not _marking_ the removed validator, won't the next execution of `update_validator_sets` override the result of this function?

N.B.: It is not totally true that we are not marking the removed validator, because the caller of this function is jailing it, but `update_validator_sets` doesn't take it into account.

So, IMO, we should remove `remove_from_validator_sets`, and modify `update_validator_sets` to remove any jailed validator from the active set.

sergio-mena: Follow up...
I see you remove the validator from _both_ sets, so I realize `update_validator_sets` won't really override the effects of this.
However, how (and when) will we reintroduce that validator back into one of the two sets?

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -0,0 +1,517 @@
...
+      var slashed_amount = vpower * slash.rate
+      //update voting power (Step 2.4 of cubic slashing)
+      //compute new total_deltas for next epoch
+      var total_offset = read_epoched_field(validators[validator_address].total_deltas, cur_epoch+1, 0)
+      validators[validator_address].total_deltas[cur_epoch+1] = total_offset - slashed_amount 
+      //update validator's voting_power and total_voting_power for next epoch
+      update_voting_power(slash.validator, cur_epoch+1)
+      update_total_voting_power(cur_epoch+1)
+    //unfrozen the validator (Step 2.5 of cubic slashing)
+    frozen[validator_address] = false
```
sergio-mena:
```suggestion
    //unfreeze the validator (Step 2.5 of cubic slashing)
    validators[validator_address].frozen = false
```

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -0,0 +1,517 @@
...
+type VotingPower uint
+
+type Validator struct {
+  consensus_key map<Epoch, Key>
+  state map<Epoch, {inactive, candidate}>
+  total_deltas map<Epoch, amount:int>
+  voting_power map<Epoch, VotingPower>
+  reward_address Addr
+  jailed bool
+  frozen bool
```
sergio-mena, line 21: From what was discussed in the last meeting with Heliax folks, I understand why `frozen` is not epoched.
However, I don't think I understand why `jailed` is not epoched. Probably has to do with the fact that we don't have the "unjailing" mechanism in the model yet

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -0,0 +1,517 @@
...
+    case duplicate_vote: return duplicate_vote_rate * voting_power
+    case ligth_client_attack: return ligth_client_attack_rate * voting_power
+    default: panic()
+}
+```
+
+```go
+end_of_epoch()
+{
+  set_validators = {val | val = slash.validator && slash in enqueued_slashes[cur_epoch]}
```
sergio-mena:
```suggestion
  var set_validators = {val | val = slash.validator && slash in enqueued_slashes[cur_epoch]}
```

sergio-mena: Not sure about this

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
sergio-mena: By `slash.validator`, did you mean `s.validator` ?

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -0,0 +1,517 @@
...
+
+```go
+//This function is called by transactions tx_withdraw_unbonds_validator and tx_withdraw_unbonds_delegator
+func withdraw(validator_address, delegator_address)
+{
+  var delunbonds = {<start,end,amount> | amount = unbonds[delegator_address][validator_address].deltas[(start, end)] > 0 && end <= cur_epoch }
+  //substract any pending slash before withdrawing
+  forall (<start,end,amount> in selfunbonds) do
+    var amount_after_slashing = amount
+    forall (slash in slashes[validator_address] s.t. start <= slash.epoch && slash.epoch <= end)
```
sergio-mena, line 262: So, what would happen if validator `validator_address`:

* doesn't have `slashes` (meaning, no punishment against it was processed -- yet -- as part of `end_of_epoch`)
* there are items in `enqueued_slashes` (for an epoch `e > cur_epoch`) that refer to `validator_address`

How is this case handled?

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
angbrav: yes, It was missing a parameter.

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
angbrav: > However, how (and when) will we reintroduce that validator back into one of the two sets?

There is no way to unjail validators at the moment, it is unspecified in Heliax's specification.

angbrav: >So, IMO, we should remove remove_from_validator_sets, and modify update_validator_sets to remove any jailed validator from the active set.

It is cosmetics. Let's first make sure it behaves as expected. We can later merge both functions if we think it is useful.

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
angbrav: No particular reason, trying to stick to the specification: in my understanding, jailed is not epoched. We have to figure out a way to deal with epoched variables that can be updated at different offsets. Once we resolve that, we could think of merging jailed (or frozen) into the variable state.

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
angbrav: These are the type of things we want to check with the TLA+ model. I believe this is related to issue https://github.com/informalsystems/partnership-heliax/issues/7, for which there is no solution currently.

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
angbrav: This is a possibility. I lean towards making delegators responsible for tracking validators though. I agree that some support from the PoS module will help delegators to avoid delegating to non-active or jailed validators, but it complicates the model. I think it is more interesting to simplify this type of logic as much as possible and focus on the most fundamental aspects of the PoS model, e.g., make sure that if a validator misbehaves and this is detected on time, we are able to slash.

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
angbrav: > As we are iterating over the previous and next epochs, and we will repeat this step for cur_epoch + 1, are we not punishing the validator multiple times for the same evidence. So for evidence from cur_epoch, the validator is punished in epochs cur_epoch -1, cur_epoch and cur_epoch + 1.

I think I made the same comment during one meeting. This is intended though. The way I currently interpret this is that any evidence **may contribute** to multiple slashes, instead of thinking a validator may be punished twice for the same evidence. So for instance, if there is evidence for a validator at epochs cur_epoch and cur_epoch + 1, the evidence at cur_epoch contributes to the slash rate of the slash computed at cur_epoch + 1 and vice-versa. Once of the goals of the slashing mechanism is to severe punishing to validators that misbehave in consecutive epochs.

angbrav: > Related to this, at the end of` curr_epoch`, we decrease the voting power (and the total voting power). But then at the end of`curr_epoch + 1`we will do the same thing taking into account an at that point invalid voting power and decreasing it again.

I think my previous answer may clarify this. Let me know if it does not.

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
sergio-mena: OK, please add a TODO in the model so we don't forget to specify unjailing (after discussing with Heliax, and understanding how they want to model unjailing)

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
sergio-mena: I agree that dealing with epoched variables that can be updated at different offsets is the biggest and most urgent of our concerns, given the ramifications it has (this is an example).

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
sergio-mena: I see. If there is no TODO in the model's text, please add one, as this has to be dealt with to have a correct model

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
sergio-mena: Yes, my understanding of what Heliax want (we discussed it in one of the last meetings) matches Manu's. I believe they are aware of the implications that @jmalicevic points out in her comment above

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
angbrav: Done

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
angbrav: I have copied the link to this comment above the `withdraw` function

angbrav:
I think everything is addressed and discussed. I am merging this and working on how to deal with epoched variables that are updated at different offsets in a different PR.

PR #17, opened 2022-08-03
Add unjailing and handles epoched variables that udpate at diff. offsets

angbrav:
This PR:
- adds unjailing logic
- handles epoched variables that update at different offsets

### DIFF
--- a/2022/Q2/artifacts/pos-model/PoS-model-cubic.md
+++ b/2022/Q2/artifacts/pos-model/PoS-model-cubic.md
@@ -2,8 +2,7 @@
 
 - There is a single token type
 - There is an initial set of active and inactive validators
-- There are always enough non-jalied, candidate validators to fulfilled the active set of validators at any given epoch
-- There is no unjailing at the moment
+- There are always enough non-jalied, candidate validators to fulfill the active set of validators at any given epoch
 
 ## Data types
 
@@ -13,13 +12,18 @@ type Key
 type Epoch uint
 type VotingPower uint
 
+type JailRecord struct {
+  is_jailed bool
+  epoch Epoch
+}
+
 type Validator struct {
   consensus_key map<Epoch, Key>
   state map<Epoch, {inactive, candidate}>
   total_deltas map<Epoch, amount:int>
   voting_power map<Epoch, VotingPower>
   reward_address Addr
-  jailed bool
+  jail_record JailRecord
   frozen bool
 }
 
@@ -27,7 +31,7 @@ type Bond struct {
   validator Addr //not used
   source Addr //not used
   deltas map<Epoch, int>
-  }
+}
 
 type Unbond struct {
   validator Addr //not used
@@ -57,9 +61,10 @@ type ValidatorSet struct {
 ```go
 pipeline_length uint
 unbonding_length uint
+min_sentence uint
 votes_per_token uint
-duplicate_vote_rate float //not used in cubic slashing
-ligth_client_attack_rate float //not used in cubic slashing
+duplicate_vote_rate float
+ligth_client_attack_rate float
 ```
 
 ## Variables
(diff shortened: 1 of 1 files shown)


2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -105,16 +115,32 @@ tx_deactivate(validator_address)
...
+    //add validator to the inactive set
+    add_validator_to_sets(validator_address, pipeline_length)
+}
+```
+
+```go
+tx_unjail(validator_address)
+{
+  //check validator is jailed and can be unjailed
+  if (is_jailed(validator_address) && (cur_epoch - validators[validator_address].jail_record.epoch > min_sentence)) then
```
tzemanovic: I think in here @sergio-mena suggested we use `cur_epoch - validators[validator_address].jail_record.epoch + pipeline_length > min_sentence)`, because we're adding the validator back to the validator sets at pipeline offset

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -88,8 +95,11 @@ tx_become_validator(validator_address, consensus_key, staking_reward_address)
     //set status to candidate and consensus key at n + pipeline_length
     validators[validator_address].consensus_key[cur_epoch+pipeline_length] = consensus_key
     validators[validator_address].state[cur_epoch+pipeline_length] = candidate
-    validators[validator_address].jailed = false
+    validators[validator_address].jail_record = JailRecord{is_jailed: false, epoch: ⊥}
     validators[validator_address].frozen = false
+    //add validator to the inactive set
+    add_validator_to_inactive(validator_address, pipeline_length)
```
tzemanovic: maybe this could use this function instead?
```suggestion
    add_validator_to_sets(validator_address, pipeline_length)
```

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
```diff
@@ -281,89 +301,152 @@ func withdraw(validator_address, delegator_address)
...
-  var slash = Slash{epoch: evidence.epoch, rate: 0, vpower_fraction: compute_vpower_fraction(evidence.type, vpower)}
+  var total_staked = read_epoched_field(validators[evidence.validator].total_deltas, evidence.epoch, 0)
+  var slash = Slash{epoch: evidence.epoch, rate: 0, stake_fraction: compute_stake_fraction(evidence.type, total_staked)}
   //enqueue slash (Step 1.1 of cubic slashing)
   append(enqueued_slashes[evidence.epoch + unbonding_length], slash)
   //jail validator (Step 1.2 of cubic slashing)
-  validators[validator_address].jailed = true
-  remove_from_validator_sets(validator_address, cur_epoch + 1)
+  validators[validator_address].jail_record = JailRecord{is_jailed: true, epoch: cur_epoch+1}
+  remove_from_validator_sets(validator_address, 1)
```
tzemanovic: (renamed)
```suggestion
  remove_validator_from_sets(validator_address, 1)
```

tzemanovic (commented):
I think when remove a validator from the validator sets their voting voting power shouldn't be included in the total voting power, so I think we should recalculate it every time we remove/add validator

2022/Q2/artifacts/pos-model/PoS-model-cubic.md
angbrav: Yes, I think so :)

angbrav:
Thanks for your comments!

> I think when remove a validator from the validator sets their voting voting power shouldn't be included in the total voting power, so I think we should recalculate it every time we remove/add validator

Yes, agree.

angbrav:
Supersede by https://github.com/informalsystems/partnership-heliax/pull/20.

PR #19, opened 2022-08-17
Reworked TLA+ model with cubic slashing

angbrav:
This PR includes major changes to the TLA+ model:
- new data types for bonds and unbonds
- totalDeltas (old bonds) is twice the size (0..2*UnbondingLength). This is important to recover the stake of a validator when evidence for an epoch < cur_epoch is found.
- adds cubic slashing
- fix for https://github.com/informalsystems/partnership-heliax/issues/7
- fix for https://github.com/informalsystems/partnership-heliax/issues/21

### DIFF
--- a/2022/Q3/artifacts/PoS-tla/MC_Staking.tla
+++ b/2022/Q3/artifacts/PoS-tla/MC_Staking.tla
@@ -8,11 +8,11 @@ UserAddrs == { "user2", "user3", "val1", "val2"}
 \* Set of two validators.
 ValidatorAddrs == {"val1", "val2"}
 
-PipelineLength == 2
+PipelineLength == 1
 
-UnbondingLength == 4
+UnbondingLength == 1
 
-TxsEpoch == 2
+TxsEpoch == 1
 
 VARIABLES
     \* Token balance for every account.
@@ -43,13 +43,9 @@ VARIABLES
     \*
     \* @type: ENQUEUEDSLASHES;
     enqueuedSlashes,
-    \* Set of jailed validators
-    \*
-    \* @type: Set(ADDR);
-    jailedValidators,
     \* Set of frozen validators
     \*
-    \* @type: Set(ADDR);
+    \* @type: FROZEN;
     frozenValidators
 
 \* Variables that model transactions, not the state machine.
@@ -93,9 +89,21 @@ NoFiveTransactions ==
 NoTenTransactions ==
     nextTxId < 10 \/ failed
 
+\* No successful withdrawing. Use it to produce a counterexample.
+NoSuccessfulWithdraw ==
+    LET Example ==
+        /\ lastTx.tag = "withdraw"
+        /\ lastTx.value > 0
+    IN
+    ~Example
+
 \* No withdrawing. Use it to produce a counterexample.
 NoWithdraw ==
     lastTx.tag /= "withdraw"
+
+\* No evidence. Use it to produce a counterexample.
+NoEvidence ==
+    lastTx.tag /= "evidence"
 
 \* From Chris
 \* Try to capture that for a group of validators with total voting power X at
--- a/2022/Q3/artifacts/PoS-tla/MC_Staking.tla
+++ b/2022/Q3/artifacts/PoS-tla/MC_Staking.tla
@@ -81,29 +81,30 @@ INSTANCE Staking
 
 \* invariants to check and break the system
 
-\* a counterexample to this invariant will produce five transactions,
-NoFiveTransactions ==
-    nextTxId < 5 \/ failed
-
-\* a counterexample to this invariant will produce ten transactions,
-NoTenTransactions ==
-    nextTxId < 10 \/ failed
-
 \* No successful withdrawing. Use it to produce a counterexample.
 NoSuccessfulWithdraw ==
     LET Example ==
         /\ lastTx.tag = "withdraw"
         /\ lastTx.value > 0
+        /\ ~lastTx.fail
     IN
     ~Example
 
 \* No withdrawing. Use it to produce a counterexample.
 NoWithdraw ==
-    lastTx.tag /= "withdraw"
+    LET Example ==
+        /\ lastTx.tag = "withdraw"
+        /\ ~lastTx.fail
+    IN
+    ~Example
 
 \* No evidence. Use it to produce a counterexample.
 NoEvidence ==
-    lastTx.tag /= "evidence"
+    LET Example ==
+        /\ lastTx.tag = "evidence"
+        /\ ~lastTx.fail
+    IN
+    ~Example
 
 \* From Chris
 \* Try to capture that for a group of validators with total voting power X at
--- a/2022/Q3/artifacts/PoS-tla/MC_Staking.tla
+++ b/2022/Q3/artifacts/PoS-tla/MC_Staking.tla
@@ -31,6 +31,10 @@ VARIABLES
     \*
     \* @type: TOTALDELTAS;
     totalDeltas,
+    \* Stake unbonded from a given validator at a given epoch.
+    \*
+    \* @type: TOTALUNBONDED;
+    totalUnbonded,
     \* PoS special account
     \*
     \* @type: Int;
--- a/2022/Q3/artifacts/PoS-tla/MC_Staking.tla
+++ b/2022/Q3/artifacts/PoS-tla/MC_Staking.tla
@@ -10,27 +10,47 @@ ValidatorAddrs == {"val1", "val2"}
 
 PipelineLength == 2
 
-UnbondingLength == 6
+UnbondingLength == 4
 
-TxsEpoch == 4
+TxsEpoch == 2
 
 VARIABLES
-    \* Coin balance for every Cosmos account.
+    \* Token balance for every account.
     \*
     \* @type: BALANCE;
     balanceOf,
-    \* Balance of unbonded coins that cannot be used during the bonding period.
+    \* Balance of unbonded tokens that cannot be used during the bonding period.
     \*
     \* @type: UNBONDED;
     unbonded,
-    \* Coins that are delegated to Validator.
-    \*
-    \* @type: DELEGATED;
-    delegated,
-    \* Voting power of the validator.
+    \* Tokens that are delegated to a validator.
     \*
     \* @type: BONDED;
-    bonded
+    bonded,
+    \* Stake at a given validator.
+    \*
+    \* @type: TOTALDELTAS;
+    totalDeltas,
+    \* PoS special account
+    \*
+    \* @type: Int;
+    posAccount,
+    \* Slashes
+    \*
+    \* @type: SLASHES;
+    slashes,
+    \* Enqueued slashes
+    \*
+    \* @type: ENQUEUEDSLASHES;
+    enqueuedSlashes,
+    \* Set of jailed validators
+    \*
+    \* @type: Set(ADDR);
+    jailedValidators,
+    \* Set of frozen validators
+    \*
+    \* @type: Set(ADDR);
+    frozenValidators
 
 \* Variables that model transactions, not the state machine.
 VARIABLES    
@@ -47,6 +67,15 @@ VARIABLES
     \* A serial number used to identify epochs
     \* @type: Int;
     epoch,
+    \* A serial number used to identify bonds
+    \* @type: Int;
+    idBonds,
+    \* A serial number used to identify unbonds
+    \* @type: Int;
+    idUnbonds,
+    \* A serial number used to identify slashes
+    \* @type: Int;
+    idSlashes,
     \* Whether at least one transaction has failed
     \* @type: Bool;
     failed
@@ -56,6 +85,10 @@ INSTANCE Staking
 
 \* invariants to check and break the system
 
+\* a counterexample to this invariant will produce five transactions,
+NoFiveTransactions ==
+    nextTxId < 5 \/ failed
+
 \* a counterexample to this invariant will produce ten transactions,
 NoTenTransactions ==
     nextTxId < 10 \/ failed
@@ -64,14 +97,34 @@ NoTenTransactions ==
 NoWithdraw ==
     lastTx.tag /= "withdraw"
 
-ValDelegatedFold(set, val) == LET SumDelegated(p,q) == p + delegated[1, q, val]
+\* From Chris
+\* Try to capture that for a group of validators with total voting power X at
+\* a particular height, the proof-of-stake model provides the ability to look up
+\* bonds contributing to that voting power (for the unbonding period) and slash
+\* a proportional amount of stake to X (subject to limitations e.g. repeated infractions)
+
+\* Invariant #1
+\* the validator's voting power at a given epoch is equal to the total amount of
+\* tokens delegated to that validator
+
+(* ValDelegatedFold(set, val) == LET SumDelegated(p,q) == p + delegated[1, q, val]
                               IN ApaFoldSet( SumDelegated, 0, set )
 
 VotingpowerDelagations ==
     \A val \in ValidatorAddrs:
-    ValDelegatedFold(UserAddrs, val) = bonded[1, val]
+    ValDelegatedFold(UserAddrs, val) = bonded[1, val] *)
+
+\* Invariant #2
+\* the user balance is always greater or equal to zero
 
 BalanceAlwaysPositive == 
     \A user \in UserAddrs: balanceOf[user] >= 0
+
+
+\* Invariant #3
+\* posAccount is always greater or equal to zero
+
+PoSAccountAlwaysPositive == 
+    posAccount >= 0
 
 ===============================================================================
--- a/2022/Q3/artifacts/PoS-tla/Staking.tla
+++ b/2022/Q3/artifacts/PoS-tla/Staking.tla
@@ -60,13 +60,9 @@ VARIABLES
     \*
     \* @type: ENQUEUEDSLASHES;
     enqueuedSlashes,
-    \* Set of jailed validators
-    \*
-    \* @type: Set(ADDR);
-    jailedValidators,
     \* Set of frozen validators
     \*
-    \* @type: Set(ADDR);
+    \* @type: FROZEN;
     frozenValidators
 
 
@@ -124,18 +120,13 @@ Init ==
                   THEN { [ id |-> 0, epoch |-> 1, amount |-> INIT_VALIDATOR_STAKE ] }
                   ELSE {}
                 ]
-    /\ totalDeltas = [ n \in 1..2*UnbondingLength, a \in ValidatorAddrs |-> 
-                       IF n <= UnbondingLength
-                       THEN 0
-                       ELSE INIT_VALIDATOR_STAKE 
-                     ]
+    /\ totalDeltas = [ n \in 0..2*UnbondingLength, a \in ValidatorAddrs |-> INIT_VALIDATOR_STAKE ]
     /\ slashes = [ b \in ValidatorAddrs |-> {} ]
-    /\ enqueuedSlashes = [ n \in 0..UnbondingLength, a \in ValidatorAddrs |-> {}] 
-    /\ jailedValidators = {}
-    /\ frozenValidators = {}
+    /\ enqueuedSlashes = [ n \in -1..UnbondingLength, a \in ValidatorAddrs |-> {}] 
+    /\ frozenValidators = [ n \in 0..UnbondingLength |-> {} ]
     /\ posAccount = Cardinality(ValidatorAddrs) * INIT_VALIDATOR_STAKE
     /\ nextTxId = 0
-    /\ epoch = 1
+    /\ epoch = UnbondingLength + 1
     /\ txCounter = 0
     /\ idBonds = 1
     /\ idUnbonds = 1
@@ -154,7 +145,7 @@ Delegate(sender, validator, amount) ==
     /\ failed' = (fail \/ failed)
     /\ nextTxId' = nextTxId + 1
     /\ txCounter' = txCounter + 1
-    /\ UNCHANGED <<epoch, unbonded, idUnbonds, slashes, idSlashes, enqueuedSlashes, jailedValidators, frozenValidators>>
+    /\ UNCHANGED <<epoch, unbonded, idUnbonds, slashes, idSlashes, enqueuedSlashes, frozenValidators>>
     /\  IF fail
         THEN
           UNCHANGED <<balanceOf, posAccount, bonded, idBonds, totalDeltas>>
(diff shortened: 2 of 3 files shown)


2022/Q3/artifacts/PoS-tla/MC_Staking.tla
```diff
@@ -56,22 +81,58 @@ INSTANCE Staking
...
+
 \* a counterexample to this invariant will produce ten transactions,
 NoTenTransactions ==
     nextTxId < 10 \/ failed
 
+\* No successful withdrawing. Use it to produce a counterexample.
+NoSuccessfulWithdraw ==
+    LET Example ==
+        /\ lastTx.tag = "withdraw"
+        /\ lastTx.value > 0
```
Kukovec, line 92: I'd recommend to either add
```suggestion
        /\ lastTx.tag = "withdraw"
        /\ lastTx.value > 0
        /\ ~lastTx.fail
```
or make a note that this should only be used with `NextNoFail`. Same for the rest of the invariants.

2022/Q3/artifacts/PoS-tla/MC_Staking.tla
```diff
@@ -56,22 +81,58 @@ INSTANCE Staking
 
 \* invariants to check and break the system
 
+\* a counterexample to this invariant will produce five transactions,
+NoFiveTransactions ==
+    nextTxId < 5 \/ failed
```
Kukovec: Just a note, the `failed` flag may be set at any point before the 5th transition, not necessarily exactly on the 5th. If you get a trace of length 5 and the fail happens after the 1st transition, the other 4 aren't really meaningful. To control the exact failure point you can use `~failed /\ failed'` or trace invariants.

2022/Q3/artifacts/PoS-tla/Staking.tla
```diff
@@ -78,32 +101,39 @@ MAX_UINT == 2^(256 - 60) - 1
...
-        ELSE 0
-       ]
-    /\ bonded = [ n \in 1..UnbondingLength, a \in ValidatorAddrs |-> INIT_VALIDATOR_STAKE ]
+                     IF a \in ValidatorAddrs
+                     THEN INITIAL_SUPPLY - INIT_VALIDATOR_STAKE
+                     ELSE INITIAL_SUPPLY
+                   ]
+    /\ unbonded = [ a \in UserAddrs, b \in ValidatorAddrs |-> {} ]
+    /\ bonded = [ a \in UserAddrs, b \in ValidatorAddrs |->
+                  IF a \in ValidatorAddrs /\ a = b
```
Kukovec: `b \in ValidatorAddrs` by definition, so `a = b` implies `a \in ValidatorAddrs`.
```suggestion
                  IF a = b
```

2022/Q3/artifacts/PoS-tla/Staking.tla
```diff
@@ -113,104 +143,257 @@ Delegate(sender, validator, amount) ==
...
+          /\ posAccount' = posAccount + amount
+          /\ bonded' = [ bonded EXCEPT ![sender, validator] = @ \union {[ id |-> idBonds,
+                                                                          epoch |-> epoch + PipelineLength,
+                                                                          amount |-> amount ]}]
+          /\ idBonds' = idBonds + 1
+          \* updates totalDeltas from PipelineLength to UnbondingLength 
+          /\ totalDeltas' = [ n \in 0..2*UnbondingLength, val \in ValidatorAddrs |-> 
+                              IF n >= UnbondingLength + PipelineLength /\ val = validator
+                              THEN totalDeltas[n, val] + amount
+                              ELSE totalDeltas[n, val]
```
Kukovec: You can do fewer independent evaluations of `totalDeltas[n, val]` by pulling it out of ITE:
```suggestion
                              totalDeltas[n, val] + 
                              IF n >= UnbondingLength + PipelineLength /\ val = validator
                              THEN amount
                              ELSE 0
```
or
```suggestion
                              LET deltas == totalDeltas[n, val] IN
                              IF n >= UnbondingLength + PipelineLength /\ val = validator
                              THEN deltas + amount
                              ELSE deltas
```

2022/Q3/artifacts/PoS-tla/Staking.tla
```diff
@@ -113,104 +143,257 @@ Delegate(sender, validator, amount) ==
...
+                                          THEN [ remaining |-> record.remaining - amount,
+                                                 set |-> record.set \union {[id |-> record.id, amount |-> bond.amount, start |-> bond.epoch, end |-> e]},
+                                                 id |-> record.id + 1 ]
+                                          ELSE [ remaining |-> 0,
+                                                 set |-> record.set \union {[id |-> record.id, amount |-> record.remaining, start |-> bond.epoch, end |-> e]},
+                                                 id |-> record.id + 1 ]    
+                                         ELSE [ remaining |-> 0,
+                                                set |-> record.set,
+                                                id |-> record.id ]
+                                        IN ApaFoldSet(F, [ remaining |-> amount, set |-> {}, id |-> idUnbonds], setBonds)
```
Kukovec: I feel like this chunk is not correct.
You initialized `ApaFoldSet` with the value `[ remaining |-> amount, set |-> {}, id |-> idUnbonds]`, therefore, `record.remaining = amount` in the first iteration. In particular, this makes the `IF record.remaining > amount` check immediately false, so this stutters after 1 step with 
```
[ remaining |-> 0,
  set |-> {
    [ id |-> idUnbonds, 
      amount |-> amount, 
      start |-> bond.epoch, 
      end |-> e]
  },
  id |-> idUnbonds + 1 
]    
```
being the final value, for one arbitrarily selected `bond \in setBonds`.
The core issue is the seeded `remaining |-> amount`. It is also probably the case that you don't need to fold over records for this, all you need is the`set` value, because IIUC, `record.remaining`  is always reduced by exactly `amount`, so you can calculate the number of iterations (`min(|setBonds|, ceil(initialAmount / amount)`), which is also the final `id` delta.

angbrav: Good catch! This is a typo. it should be `IF record.remaining > bond.amount`. I've fixed it.

angbrav:
Thanks for the comments @Kukovec!

2022/Q3/artifacts/PoS-tla/Staking.tla
```diff
@@ -113,104 +143,257 @@ Delegate(sender, validator, amount) ==
...
+                                          THEN [ remaining |-> record.remaining - bond.amount,
+                                                 set |-> record.set \union {[id |-> record.id, amount |-> bond.amount, start |-> bond.epoch, end |-> e]},
+                                                 id |-> record.id + 1 ]
+                                          ELSE [ remaining |-> 0,
+                                                 set |-> record.set \union {[id |-> record.id, amount |-> record.remaining, start |-> bond.epoch, end |-> e]},
+                                                 id |-> record.id + 1 ]    
+                                         ELSE [ remaining |-> 0,
+                                                set |-> record.set,
+                                                id |-> record.id ]
+                                        IN ApaFoldSet(F, [ remaining |-> totalAmount, set |-> {}, id |-> idUnbonds], setBonds)
```
Kukovec: You can actually save 1 ITE by doing:
```suggestion
ComputedUnbonds(setBonds, totalAmount, e) == LET 
                                         \* @type: ([ remaining: Int, set: Set(UNBOND), id: Int ], BOND) => [ remaining: Int, set: Set(UNBOND), id: Int ];
                                         F(record, bond) == 
                                         IF record.remaining > 0
                                         THEN 
                                          LET min == Min(record.remaining, bond.amount) 
                                          IN
                                          [ remaining |-> record.remaining - min,
                                            set |-> record.set \union {[id |-> record.id, amount |-> min, start |-> bond.epoch, end |-> e]},
                                            id |-> record.id + 1 ]   
                                         ELSE [ remaining |-> 0,
                                                set |-> record.set,
                                                id |-> record.id ]
                                        IN ApaFoldSet(F, [ remaining |-> totalAmount, set |-> {}, id |-> idUnbonds], setBonds)
```

2022/Q3/artifacts/PoS-tla/MC_Staking.tla
```diff
@@ -8,29 +8,45 @@ UserAddrs == { "user2", "user3", "val1", "val2"}
...
 ValidatorAddrs == {"val1", "val2"}
 
-PipelineLength == 2
+PipelineLength == 1
 
-UnbondingLength == 6
+UnbondingLength == 1
 
-TxsEpoch == 4
+TxsEpoch == 1
```
sergio-mena, line 15: Reason for these changes? Debugging?

2022/Q3/artifacts/PoS-tla/Staking.tla
```diff
@@ -1,17 +1,15 @@
...
+ * Modeling the Anoma's PoS module.
  *
  * Simplifications:
- *   - We assume only one validator to be delegated to.
- *   - Rewards are not specified.
- *   - Jailing is not specified.
- *   - Slashing and evidence handling is not specified.
+ *   - Withdraws take a validator as argument
+ *   - Rewards and unjailing are not specified
+ *   - Each validator can misbehave at most once per execution
```
sergio-mena, line 8: While revisiting #7 (Tomas's solution) I had a problematic example with two infractions committed in consecutive epochs

2022/Q3/artifacts/PoS-tla/Staking.tla
```diff
@@ -78,32 +101,49 @@ MAX_UINT == 2^(256 - 60) - 1
 INITIAL_SUPPLY == 10^(9+18)
 
 \* the number of tokens the validator has staked
+\* @type: () => Int;
 INIT_VALIDATOR_STAKE == 1000000000000000000000
 
 \* the amount of voting power per token
```
sergio-mena, line 111: Is this comment still valid?

2022/Q3/artifacts/PoS-tla/Staking.tla
```diff
@@ -78,32 +101,49 @@ MAX_UINT == 2^(256 - 60) - 1
...
+                  ELSE {}
+                ]
+    /\ totalDeltas = [ n \in 0..2*UnbondingLength, a \in ValidatorAddrs |-> INIT_VALIDATOR_STAKE ]
+    /\ slashes = [ b \in ValidatorAddrs |-> {} ]
+    /\ enqueuedSlashes = [ n \in -1..UnbondingLength, a \in ValidatorAddrs |-> {}] 
+    /\ frozenValidators = [ n \in 0..UnbondingLength |-> {} ]
+    /\ posAccount = Cardinality(ValidatorAddrs) * INIT_VALIDATOR_STAKE
     /\ nextTxId = 0
-    /\ epoch = 1
+    /\ epoch = UnbondingLength + 1
```
sergio-mena, line 145: Reason for this change?

2022/Q3/artifacts/PoS-tla/Staking.tla
```diff
@@ -78,32 +101,49 @@ MAX_UINT == 2^(256 - 60) - 1
...
     /\ txCounter = 0
+    /\ idBonds = 1
+    /\ idUnbonds = 1
+    /\ idSlashes = 1
     /\ lastTx = [ id |-> 0, tag |-> "None", fail |-> FALSE ]
     /\ failed = FALSE
 
-
 \* delegate `amount` tokens to a validator
 Delegate(sender, validator, amount) ==
```
sergio-mena, line 154: If "sender" can be a validator, why do we need "Init" to be so complex? The state Init is trying to set up can be reached with a sequence of "Delegate", can't it?

2022/Q3/artifacts/PoS-tla/Staking.tla
```diff
@@ -113,104 +153,245 @@ Delegate(sender, validator, amount) ==
...
-
-\* unbond `amount` tokens from a validator
+          /\ posAccount' = posAccount + amount
+          /\ bonded' = [ bonded EXCEPT ![sender, validator] = @ \union {[ id |-> idBonds,
+                                                                          epoch |-> epoch + PipelineLength,
+                                                                          amount |-> amount ]}]
+          /\ idBonds' = idBonds + 1
+          \* updates totalDeltas from PipelineLength to UnbondingLength 
+          /\ totalDeltas' = [ n \in 0..2*UnbondingLength, val \in ValidatorAddrs |-> totalDeltas[n, val] + 
+                              IF n >= UnbondingLength + PipelineLength /\ val = validator
```
sergio-mena, line 179: `n >= UnbondingLength + PipelineLength` can you explain this condition?

2022/Q3/artifacts/PoS-tla/Staking.tla
```diff
@@ -113,104 +153,245 @@ Delegate(sender, validator, amount) ==
...
+\* @type: (ADDR, ADDR) => Int;
+TotalBonds(sender, validator) == LET 
+                                 \* @type: (Int, BOND) => Int;
+                                 F(total, bond) == total + bond.amount
+                                 IN ApaFoldSet(F, 0, bonded[sender, validator])
+
+\* @type: (Set(BOND), Int, Int) => [ remaining: Int, set: Set(UNBOND), id: Int ];
+ComputedUnbonds(setBonds, totalAmount, e) == LET 
+                                         \* @type: ([ remaining: Int, set: Set(UNBOND), id: Int ], BOND) => [ remaining: Int, set: Set(UNBOND), id: Int ];
+                                         F(record, bond) ==
```
sergio-mena, line 195: Would it be clearer to rename "bond" to "unbond" here?

2022/Q3/artifacts/PoS-tla/Staking.tla
```diff
@@ -113,104 +153,245 @@ Delegate(sender, validator, amount) ==
...
+ComputedUnbonds(setBonds, totalAmount, e) == LET 
+                                         \* @type: ([ remaining: Int, set: Set(UNBOND), id: Int ], BOND) => [ remaining: Int, set: Set(UNBOND), id: Int ];
+                                         F(record, bond) == 
+                                         IF record.remaining > 0
+                                         THEN 
+                                          LET min == Min(record.remaining, bond.amount) 
+                                          IN
+                                          [ remaining |-> record.remaining - min,
+                                            set |-> record.set \union {[id |-> record.id, amount |-> min, start |-> bond.epoch, end |-> e]},
+                                            id |-> record.id + 1 ]
```
sergio-mena: Can you clarify the goal of incrementing the id here?

2022/Q3/artifacts/PoS-tla/Staking.tla
```diff
@@ -113,104 +153,245 @@ Delegate(sender, validator, amount) ==
...
+
+(*
+* For a given unbond, it computes the amount to be withdraw after applying a set of slashes.
+*)
+\* @type: (Set(SLASH), UNBOND) => Int;
+ApplySlashes(setSlashes, unbond) == 
+    LET 
+    \* @type: (Int, SLASH) => Int;
+    F(total, slash) == total - unbond.amount*slash.finalRate
+    IN ApaFoldSet(F, unbond.amount, setSlashes)
```
sergio-mena, line 247: Can't this become negative?

2022/Q3/artifacts/PoS-tla/Staking.tla
```diff
@@ -113,104 +153,245 @@ Delegate(sender, validator, amount) ==
...
+      EndOfEpoch
     ELSE
       \E sender \in UserAddrs:
       \E validator \in ValidatorAddrs:
       \E amount \in Int:
         /\ amount <= MAX_UINT
-        /\ \/ Delegate(sender, validator, amount)
+        \* this guarantees that a validator does not misbehave more than one in an UnbondingLength period.
+        /\ \/ IF validator \notin frozenValidators[0]
+              THEN
```
sergio-mena, line 430: Why requiring this here? Wouldn't it be cleaner to do it inside evidence, just as is done in line 207?

sergio-mena (commented):
A few comments, some of them are actually clarifying questions.
The model is now advanced!!!

2022/Q3/artifacts/PoS-tla/MC_Staking.tla
angbrav: The checker explores traces of max. 10 transitions (this is configurable but a higher number would probably be too much): a transition being a transaction. If we want things to actually happen, we need to use very small numbers for those configurable parameters. 1 it is probably too much, but we cannot really go much higher.

2022/Q3/artifacts/PoS-tla/Staking.tla
angbrav: Can you please comment on the issue?

2022/Q3/artifacts/PoS-tla/Staking.tla
angbrav: Do you mean the type annotation? I think we can remove it, but I guess it is harmless.

2022/Q3/artifacts/PoS-tla/Staking.tla
angbrav: We now keep the stake of past epochs in total_deltas, as we need it when computing slashes. If we start in epoch 1, then it is kind of weird semantically to keep the stake of negative epochs.

2022/Q3/artifacts/PoS-tla/Staking.tla
angbrav: can you be more specific? which part of Init seems complex to you? Are you referring to the fact that validators have a number of tokens bonded from the beginning?

2022/Q3/artifacts/PoS-tla/Staking.tla
angbrav: index unbonding_length of total_deltas is cur_epoch. Then we want to update total deltas starting from cur_epoch+pipeline_lenght = unbonding_length+pipeline_lenght

2022/Q3/artifacts/PoS-tla/Staking.tla
angbrav: Each unbond (the same for slashes and bonds) is uniquely identified by the id. It is just a way of transforming a set into a multiset or bag.

2022/Q3/artifacts/PoS-tla/Staking.tla
angbrav: Do you mean the bond in `F(record, bond)`?

2022/Q3/artifacts/PoS-tla/Staking.tla
angbrav: Either way works. Maybe I am wrong, but I think this way we do not waste transitions (which are very precious).

2022/Q3/artifacts/PoS-tla/Staking.tla
angbrav: It can if the validator misbehaves multiple times. Let's reason about it whenever we integrate the fix to #7

2022/Q3/artifacts/PoS-tla/Staking.tla
sergio-mena: OK. But let's discuss synchronously first, so my comment can be meaningful

2022/Q3/artifacts/PoS-tla/Staking.tla
sergio-mena: The variable has changed from `VOTES_PER_TOKEN` to `SLASH_RATE`, so not sure the comment still matches

2022/Q3/artifacts/PoS-tla/Staking.tla
sergio-mena: I see, so it's a shortcut to avoid having to model the special case when there aren't enough epochs yet when you compute the slashes?

2022/Q3/artifacts/PoS-tla/Staking.tla
sergio-mena: Exactly. That can be just the first transition... so the "code" of the model would be simpler.
I know we are resource-limited, but it'd be one transition per validator only, and it would greatly simplify this part

2022/Q3/artifacts/PoS-tla/Staking.tla
sergio-mena: Wouldn't it be clearer to use `epoch` here (which is `cur_epoch` IIUC) instead of `UnbondingLength`?

2022/Q3/artifacts/PoS-tla/Staking.tla
sergio-mena: Yes, the parameter's name

2022/Q3/artifacts/PoS-tla/Staking.tla
sergio-mena: Yes, but I don't get how you _know_ that `record.id + 1` is the one you're looking for... I guess this requires some synchronous interaction

2022/Q3/artifacts/PoS-tla/Staking.tla
sergio-mena: If you use `NextNoFail` to check, wouldn't that be equivalent in that case? Since no transition can lead to a failed state?

2022/Q3/artifacts/PoS-tla/Staking.tla
sergio-mena: OK

2022/Q3/artifacts/PoS-tla/Staking.tla
angbrav: The resulting trace yes, but I guess that the checker will take longer to generate it. The latter is just a guess.

2022/Q3/artifacts/PoS-tla/Staking.tla
angbrav: Not sure I follow, it is a bond and not an unbond.

2022/Q3/artifacts/PoS-tla/Staking.tla
angbrav: That piece of code simply creates new unbonds, each unbond gets assigned a unique id.

2022/Q3/artifacts/PoS-tla/Staking.tla
angbrav: The way it is we can have "deeper" traces as the first transition could already be unbond. Nevertheless, I do not really have a strong opinion about it.

2022/Q3/artifacts/PoS-tla/Staking.tla
angbrav: I see, I'll fix it.

2022/Q3/artifacts/PoS-tla/Staking.tla
sergio-mena: Yeah, I get it now (thanks for offline explanation)

2022/Q3/artifacts/PoS-tla/Staking.tla
sergio-mena: You're right, now I get it

2022/Q3/artifacts/PoS-tla/Staking.tla
sergio-mena: After discussion, I understand the answer to my latest question is "no", because `epoch` is absolute and `UnbondingLength` represents a relative epoch (an offset), which is what these data structures are keeping.

PR #26, opened 2022-09-27
Alternative to issue #24 and missing piece for #7

### LINKED ISSUE #24: Iterate multiple times over the same set of bonds when unbonding
When a user unbonds a number of tokens = amount, the system iterates over a set a bonds and creates an unbond for each of these bonds until we reach amount. After this process, we create a new bond with -amount. This is a problem, as if there is a second unbond, we will iterate over the same set of bonds, which are supposed to have been unbonded already. We need a solution that avoids iterating over already unbonded bonds.

We have discussed two solutions:

- Simplest solution: update the amount of each bond. This means to reduce to 0 for all but the last iterated.
We are destroying info, but seems not to be a problem because we use unbonds when slashing. **Only concern:** discrepancy between total_deltas and bonds.

- An alternative solution could be one similar to total_unbonds.

We have agreed to try the simplest solution.

angbrav:
This PR addresses two issues:

(i) It proposes an alternative to fix #24. Instead of decreasing bonds to 0 when iterating over them, this PR proposes the following: 
- As unbonds, bonds have an end epoch field that is set to `bottom` when a bond is created,
- When we iterate over bonds to create unbonds, we only iterate over the bonds with `bottom` end: those are the "yet-to-unbond" bonds.
- For each bond (while there are remaining tokens to unbond), we set `bond.end` to `cur_epoch+pipeline_lenght`.
- There is special case in which the amount of the next bond is greater than the remaining. In this case, we create one new bond with `bond.amount - remaining` with end epoch equal to `bottom`, and one bond with `remaining` amount with end at `cur_epoch+pipeline_lenght`.

(ii) It proposes a fix to the latest issue spotted in #7. We cannot update total_deltas with the unbonded amount because we may have already taken part of it when slashing the validator.


brentstone:
A couple questions from me just to help my understanding:

1. Can you clarify what kind of value `bottom` is exactly?
2. When iterating over the bonds, why do we set `bond.end` to `cur_epoch+pipeline_length` rather than `cur_epoch+pipeline_length+unbonding_length`, which is where I think we will be creating the `unbond`?

angbrav:
> A couple questions from me just to help my understanding:
> 
> 1. Can you clarify what kind of value `bottom` is exactly?

I am not sure I understand your question. It means the bond has not been unbonded.

> 2. When iterating over the bonds, why do we set `bond.end` to `cur_epoch+pipeline_length` rather than `cur_epoch+pipeline_length+unbonding_length`, which is where I think we will be creating the `unbond`?

We could also do that, I do not think there is anything wrong with it. I am using `cur_epoch+pipeline_length` because it is when `total_deltas` is updated. Thus, the end field of the bond indicates when the bonded tokens stop contributing to the validator's voting power.

brentstone:
> I am not sure I understand your question. It means the bond has not been unbonded.

I mean to ask if you envision this as an Optional type that can return `None` (e.g. in Rust)? Or can it also just be a `bool`? Once `bond.end` is set to some epoch, is that information used again? Wondering if this is merely a flag to indicate that we should not iterate over this bond in the future.

angbrav:
> I mean to ask if you envision this as an Optional type that can return `None` (e.g. in Rust)? Or can it also just be a `bool`? Once `bond.end` is set to some epoch, is that information used again? Wondering if this is merely a flag to indicate that we should not iterate over this bond in the future.

In the pseudocode model it is not, so it could be a boolean. In the TLA+ model, it simplifies things a bit to track the epoch at which the bond is unbonded. We use it in some of the invariants. For instance, this allows us to figure out the amount of tokens delegated to a validator in a given epoch `e` without having to iterate over the unbonds. If `bond.end` is just a boolean, we will have to iterate over the bonds whose `bond.epoch <= e && bond.epoch == FALSE` and then iterate over all unbonds s.t. `unbond.start <= e && unbond.end > e`. I wonder if this could be useful for you as well, or not. I guess that if you do not need to do this type of computation, duplicating this information is pointless.

I'd like to leave it as it is for now though (an integer; we could use -1 as bottom), as I suspect that it may be useful to solve issue #27  (if this is an issue after all). If it turns out that it isn't, then we can then use a boolean.

Issue #25, opened 2022-09-20
PoS invariants


tzemanovic:
Brain storming some more invariants to check in PoS model:

- Disregarding any slashes or staking rewards (we're not modeling rewards yet), an amount that is self-bonded or delegated is exactly equal to the amount that can be withdrawn back from PoS (in other words, if all the bonds are withdrawn, the PoS balance must be equal to zero)
- For a validator who gets slashed for some rate of tokens, the total slashed token amount must be equal to the sum of slashed amounts on each bond to this validator that is liable for the slash (self-bonds and delegations).
- A bond submitted at epoch `n` cannot affect voting power of a target validator before epoch `n + pipeline_len`  
- bonding and unbonding applied in the same epoch preserve composition. I.e. a bond of amount `z` submitted in an epoch `e` has the same effect as 2 bonds applied in epoch `e` of amounts `x` and `y`, where `x + y = z`
- I don't think we actually have the active/inactive validator sets in the model, so not sure if these invariant are useful or even something that we'd want to check, but here goes:
	- The length of active validator set is always <= `max_validator_slots` parameter
	- A validator account is always in either in active or inactive validator set, unless it is jailed, in which case it is in neither
	- The lowest active validator's voting power is >= greatest inactive validator's voting power
	- The validators sets use a stable sort - The order of validators in the validator sets positions is preserved when there are multiple validators with the same voting power

angbrav:
Agreed to look into the first two bullets. The third and the fourth trivially hold by construction. The fifth would require a new TLA+ model (or extending the current one).

Issue #2, opened 2022-03-08
General questions


jmalicevic:
We will group here questions that do not clearly fall into a category. 

1. **Interleaving between spec and docs**
Right now, there are many links within the spec that point to the docs. As everything is still under development this is fine, but we would like to know whether the content of matching chapters is the same? As in - is the spec a subset of the docs? (We get directed to the docs and keep reading there so if the content is not the same, we need to be more careful about going back to the spec). 

2. **Gas charges**
The gas is proportional to the byte length of both the transaction and the validity predicates. Could this lead to a situation where the user is unaware of the actual cost their validity predicate might have and that he/she can end up paying more gas for the vp than the transaction itself? If so, this can be an acceptable scenario for you and that's fine, but it might be something that should be additionally emphasised. 

3. **WASM to storage**
We were wondering how is data transfered in and out of WASM. Is the answer here: https://docs.anoma.network/v0.4.0/explore/design/ledger/wasm-vm.html

PR #28, opened 2022-09-29
PoS invariants, new fix to issue #7 and parameter to control misbehaving windows

angbrav:
This PR includes:

- PoS invariants. Including some invariants from #25 and others.
- A new fix for #7.
- An alternative design for #24.
- New parameter to control how often validators can misbehave.

### DIFF
--- a/2022/Q3/artifacts/PoS-tla/MC_Staking.tla
+++ b/2022/Q3/artifacts/PoS-tla/MC_Staking.tla
@@ -14,6 +14,8 @@ UnbondingLength == 1
 
 TxsEpoch == 1
 
+MisbehavingWindow == 100
+
 VARIABLES
     \* Token balance for every account.
     \*
@@ -54,7 +56,11 @@ VARIABLES
     \* Set of frozen validators
     \*
     \* @type: FROZEN;
-    frozenValidators
+    frozenValidators,
+    \* Set of misbehaving validators
+    \*
+    \* @type: MISBEHAVING;
+    misbehavingValidators
 
 \* Variables that model transactions, not the state machine.
 VARIABLES    
--- a/2022/Q3/artifacts/PoS-tla/MC_Staking.tla
+++ b/2022/Q3/artifacts/PoS-tla/MC_Staking.tla
@@ -289,5 +289,12 @@ TotalAccounts == LET F(sum, user) == sum + balanceOf[user]
 
 TotalAmountTokensConstant ==
     TotalAccounts + posAccount + slashPool = Cardinality(UserAddrs) * INITIAL_SUPPLY
-    
+
+(* 
+ * Invariant #8
+ * Total deltas greater or equal to zero.
+*)
+
+TotalDeltasGreater ==
+    \A val \in ValidatorAddrs: totalDeltas[UnbondingLength, val] >= 0
 ===============================================================================
--- a/2022/Q3/artifacts/PoS-tla/MC_Staking.tla
+++ b/2022/Q3/artifacts/PoS-tla/MC_Staking.tla
@@ -167,27 +167,32 @@ HighCoverage(trace) ==
 \* a proportional amount of stake to X (subject to limitations e.g. repeated infractions)
 
 \* Invariant #1
-\* the validator's voting power at a given epoch is equal to the total amount of
-\* tokens delegated to that validator
-
-(* ValDelegatedFold(set, val) == LET SumDelegated(p,q) == p + delegated[1, q, val]
-                              IN ApaFoldSet( SumDelegated, 0, set )
-
-VotingpowerDelagations ==
-    \A val \in ValidatorAddrs:
-    ValDelegatedFold(UserAddrs, val) = bonded[1, val] *)
-
-\* Invariant #2
 \* the user balance is always greater or equal to zero
 
 BalanceAlwaysPositive == 
     \A user \in UserAddrs: balanceOf[user] >= 0
 
 
-\* Invariant #3
+\* Invariant #2
 \* posAccount is always greater or equal to zero
 
 PoSAccountAlwaysPositive == 
     posAccount >= 0
+
+\* Invariant #3
+\* the validator's voting power at a given epoch is equal to the total amount of
+\* tokens delegated to that validator
+
+SumBondsUser(user, val) == LET
+                            \* @type: (Int, BOND) => Int;    
+                            F(sum, bond) == sum + bond.amount
+                           IN ApaFoldSet(F, 0, { b \in bonded[user, val]: b.epoch <= epoch })
+
+TotalSumBonds(val) == LET F(sum, user) == sum + SumBondsUser(user, val)
+                      IN ApaFoldSet(F, 0, UserAddrs)
+
+VotingpowerDelagations ==
+    \A val \in ValidatorAddrs:
+    TotalSumBonds(val) = totalDeltas[UnbondingLength, val]
 
 ===============================================================================
--- a/2022/Q3/artifacts/PoS-tla/MC_Staking.tla
+++ b/2022/Q3/artifacts/PoS-tla/MC_Staking.tla
@@ -213,7 +213,7 @@ BalanceAlwaysPositive ==
  * When we slash a validator (reduce its voting power) at the end of an epoch, we move tokens from the posAccount to
  * the slashPool. Thus, in the posAccount, we only have the bonded tokens that are withdrawable: this means, the bonded
  * tokens minus the slashsable amount. If a user is able to withdraw tokens that were contributing to the voting 
- * power of the misbehaving validator, then in a execution were all toekns are withdrawn, the posAccount should go
+ * power of the misbehaving validator, then in a execution were all tokens are withdrawn, the posAccount should go
  * negative.
 *)
 
--- a/2022/Q3/artifacts/PoS-tla/MC_Staking.tla
+++ b/2022/Q3/artifacts/PoS-tla/MC_Staking.tla
@@ -179,14 +179,14 @@ HighCoverage(trace) ==
  * a proportional amount of stake to X (subject to limitations e.g. repeated infractions)
  *
  * We can break this into two properties:
- *  (i) Given a validators with total voting power X at a particular height, the 
+ *  (i) Given a validator with total voting power X at a particular height, the 
  *      proof-of-stake model provides the ability to look up bonds contributing to
  *      that voting power.
  * (ii) If evidence is found in time and there are no repeated infractions within 
  *      an epoch window, then the proof-of-stake model guarantees slashing.
  *
  * Invariant #6 verifies property (i) for the case in which there is no slashing.
- * This is the nly case in which we have totalDeltas = bonds.
+ * This is the only case in which we have totalDeltas = bonds.
  *
  * Invariant #2 verifies property (ii).
 *)
--- a/2022/Q3/artifacts/PoS-tla/MC_Staking.tla
+++ b/2022/Q3/artifacts/PoS-tla/MC_Staking.tla
@@ -39,6 +39,10 @@ VARIABLES
     \*
     \* @type: Int;
     posAccount,
+    \* Slashed special account
+    \*
+    \* @type: Int;
+    slashPool,
     \* Slashes
     \*
     \* @type: SLASHES;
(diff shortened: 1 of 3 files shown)


2022/Q3/artifacts/PoS-tla/MC_Staking.tla
```diff
@@ -111,14 +121,15 @@ NoEvidence ==
...
 (* 
-* It generates a trace with high coverage
+* It generates a trace with high coverage.
 *
-* Use it with UnbondingLength and PipelineLength equal to 1 and 
+* Use it with:
 * UserAddrs == { "user2", "user3", "val1", "val2"}
 * ValidatorAddrs == {"val1", "val2"}
 * PipelineLength == 1
 * UnbondingLength == 1
```
sergio-mena, line 130: But those are not realistic operating conditions

2022/Q3/artifacts/PoS-tla/MC_Staking.tla
```diff
@@ -160,34 +171,136 @@ HighCoverage(trace) ==
...
-\* a proportional amount of stake to X (subject to limitations e.g. repeated infractions)
+(* 
+ * From Chris:
+ * Try to capture that for a group of validators with total voting power X at
+ * a particular height, the proof-of-stake model provides the ability to look up
+ * bonds contributing to that voting power (for the unbonding period) and slash
+ * a proportional amount of stake to X (subject to limitations e.g. repeated infractions)
+ *
+ * We can break this into two properties:
+ *  (i) Given a validators with total voting power X at a particular height, the
```
sergio-mena: "a validator" or "a validator set" ?

2022/Q3/artifacts/PoS-tla/MC_Staking.tla
```diff
@@ -160,34 +171,136 @@ HighCoverage(trace) ==
...
+ * Try to capture that for a group of validators with total voting power X at
+ * a particular height, the proof-of-stake model provides the ability to look up
+ * bonds contributing to that voting power (for the unbonding period) and slash
+ * a proportional amount of stake to X (subject to limitations e.g. repeated infractions)
+ *
+ * We can break this into two properties:
+ *  (i) Given a validators with total voting power X at a particular height, the 
+ *      proof-of-stake model provides the ability to look up bonds contributing to
+ *      that voting power.
+ * (ii) If evidence is found in time and there are no repeated infractions within
```
sergio-mena, line 185: Why do you need to constrain the condition to "no repeated infractions" ?

2022/Q3/artifacts/PoS-tla/MC_Staking.tla
```diff
@@ -160,34 +171,136 @@ HighCoverage(trace) ==
...
+ *
+ * We can break this into two properties:
+ *  (i) Given a validators with total voting power X at a particular height, the 
+ *      proof-of-stake model provides the ability to look up bonds contributing to
+ *      that voting power.
+ * (ii) If evidence is found in time and there are no repeated infractions within 
+ *      an epoch window, then the proof-of-stake model guarantees slashing.
+ *
+ * Invariant #6 verifies property (i) for the case in which there is no slashing.
+ * This is the nly case in which we have totalDeltas = bonds.
```
sergio-mena:
```suggestion
 * This is the only case in which we have totalDeltas = bonds.
```

2022/Q3/artifacts/PoS-tla/MC_Staking.tla
```diff
@@ -160,34 +171,136 @@ HighCoverage(trace) ==
...
+ * a proportional amount of stake to X (subject to limitations e.g. repeated infractions)
+ *
+ * We can break this into two properties:
+ *  (i) Given a validators with total voting power X at a particular height, the 
+ *      proof-of-stake model provides the ability to look up bonds contributing to
+ *      that voting power.
+ * (ii) If evidence is found in time and there are no repeated infractions within 
+ *      an epoch window, then the proof-of-stake model guarantees slashing.
+ *
+ * Invariant #6 verifies property (i) for the case in which there is no slashing.
```
sergio-mena, line 188: Can we extend that invariant (or write another one) to also cover the case where there are slashes? (which is the tricky one)?

2022/Q3/artifacts/PoS-tla/MC_Staking.tla
```diff
@@ -160,34 +171,136 @@ HighCoverage(trace) ==
...
+ * This implies the following:
+ * Assume that a validator misbehaves at epoch e and that evidence is found at an epoch e’ <= e + unbonding_length.
+ * Assume also that the validator only misbehaves once. Then, any user whose bonds (including the validators self-bonds)
+ * were contributing to the misbehaving validator's voting power are slashed when withdrawing any of those bonds.
+ *
+ * The logic is the following:
+ * When we slash a validator (reduce its voting power) at the end of an epoch, we move tokens from the posAccount to
+ * the slashPool. Thus, in the posAccount, we only have the bonded tokens that are withdrawable: this means, the bonded
+ * tokens minus the slashsable amount. If a user is able to withdraw tokens that were contributing to the voting 
+ * power of the misbehaving validator, then in a execution were all toekns are withdrawn, the posAccount should go
```
sergio-mena:
```suggestion
 * power of the misbehaving validator, then in a execution were all tokens are withdrawn, the posAccount should go
```

2022/Q3/artifacts/PoS-tla/MC_Staking.tla
```diff
@@ -160,34 +171,136 @@ HighCoverage(trace) ==
...
+ * a particular height, the proof-of-stake model provides the ability to look up
+ * bonds contributing to that voting power (for the unbonding period) and slash
+ * a proportional amount of stake to X (subject to limitations e.g. repeated infractions)
+ *
+ * We can break this into two properties:
+ *  (i) Given a validators with total voting power X at a particular height, the 
+ *      proof-of-stake model provides the ability to look up bonds contributing to
+ *      that voting power.
+ * (ii) If evidence is found in time and there are no repeated infractions within 
+ *      an epoch window, then the proof-of-stake model guarantees slashing.
```
sergio-mena, line 186: "the proof-of-stake model guarantees slashing"
IIUC, invariant #2 guarantees that the slashing is properly computed. But it doesn't guarantee that (some) slashing takes place.
In other works. If I set all slashing to 0 in the model, will invariant#2 be violated?

2022/Q3/artifacts/PoS-tla/MC_Staking.tla
```diff
@@ -160,34 +171,136 @@ HighCoverage(trace) ==
...
+
+(* 
+ * Invariant #3
+ * First bullet from https://github.com/informalsystems/partnership-heliax/issues/25
+ *
+ * The amount that is self-bonded and delegated tokens is exactly equal to the amount that can be withdrawn back from PoS.
+ * In other words, if all the bonds are withdrawn, the PoS balance must be equal to zero.
+*)
+
+posAccountZero ==
```
sergio-mena, line 243: [Suggestion]
To better identify the actual invariants (distinguish them from auxiliary rules), we might want to prepend their name with "Inv"... or prepend the auxiliary with "Aux"

2022/Q3/artifacts/PoS-tla/MC_Staking.tla
```diff
@@ -160,34 +171,136 @@ HighCoverage(trace) ==
...
+    bondsWithOpenEnd = {} /\ foldUnbonds = {} => posAccount = 0
```
sergio-mena, line 244: (try to understand) So, when an unbond is complete, we delete the corresponding "unbond"?

sergio-mena: Further to this. Since the premise of the invariant is quite strong, I'm concerned it's working because we never get the premise true.
To double check, would it be possible to "require the premise" (by checking it negated), in order to obtain a trace, and then verify in that trace that `posAccount` is indeed 0?

2022/Q3/artifacts/PoS-tla/MC_Staking.tla
```diff
@@ -160,34 +171,136 @@ HighCoverage(trace) ==
...
+
+\* Auxiliary functions for Invariants #4, #5 and #6.            
+
+SumBondsUser(user, val) == LET
```
sergio-mena, line 248: Since `SumBondsUser` is only used as part of `TotalSumBonds`, I understand you splitted them for readability.
However, aren't we splitting a single loop into many (affecting efficiency)?

2022/Q3/artifacts/PoS-tla/MC_Staking.tla
```diff
@@ -160,34 +171,136 @@ HighCoverage(trace) ==
...
+ * The validator's voting power at a given epoch is not equal to the total amount of
+ * tokens delegated to that validator due to slashing.
+ * The follwoing invariant is used to create a counterexample.
+*)
 
-BalanceAlwaysPositive == 
-    \A user \in UserAddrs: balanceOf[user] >= 0
+VotingpowerNotEqualDelagations ==
+    \A val \in ValidatorAddrs:
+    TotalSumBonds(val) = totalDeltas[UnbondingLength, val]
```
sergio-mena, line 277: Shouldn't you also require at least one slash in the invariant?

2022/Q3/artifacts/PoS-tla/Staking.tla
```diff
@@ -205,14 +220,27 @@ ComputedUnbonds(setBonds, totalAmount, e) == LET
...
+                                        IN ApaFoldSet(F, [ remaining |-> totalAmount, unbonds |-> {}, bondsToRemove |-> {}, bondsToAdd |-> {}, id |-> idUnbonds], setBonds)
+
+\* @type: (Int, Set(SLASH)) => Int;
+BondAfterSlashing(amount, setSlashes) == LET
+                                         \* @type: (Int, SLASH) => Int;
+                                         F(total, slash) == total + (amount - amount*slash.finalRate)
+                                         IN ApaFoldSet(F, 0, setSlashes)
+
+\* @type: (ADDR, Set(BOND)) => Int;
+ComputeTakeTotalDeltas(val, setBonds) == LET
```
sergio-mena, line 239: I think I understand what this is doing, but I get lost when trying to related to the name (`ComputeTakeTotalDeltas`)

sergio-mena (commented):
Good to see all these invariants being checked!

2022/Q3/artifacts/PoS-tla/MC_Staking.tla
angbrav: a validator

2022/Q3/artifacts/PoS-tla/MC_Staking.tla
angbrav: with repeated infractions, there might not be tokens to slash (all of them have been slashed already due to previous infractions).

2022/Q3/artifacts/PoS-tla/MC_Staking.tla
angbrav: I am not sure I am following. With slashing, there will be no match between total_deltas and bonds (bonds are not updated when slashing). Invariant 4 is similar to this one but with slashing, but in that case sum(bonds) >= total_deltas because of the above reason.

2022/Q3/artifacts/PoS-tla/MC_Staking.tla
angbrav: > (try to understand) So, when an unbond is complete, we delete the corresponding "unbond"?

Yes

> Further to this. Since the premise of the invariant is quite strong, I'm concerned it's working because we never get the premise true.
> To double check, would it be possible to "require the premise" (by checking it negated), in order to obtain a trace, and then verify in that trace that `posAccount` is indeed 0?

Now it may be the case, as I am using the `simulate` feature, but when we use `check` that should not be a concern.

2022/Q3/artifacts/PoS-tla/MC_Staking.tla
angbrav: `SumBondsUser` iterates over users and `TotalSumBonds` over the bonds of a given validator and user, so I think it is not a single loop.

2022/Q3/artifacts/PoS-tla/Staking.tla
angbrav: Agree, it is a bad name. Any suggestion? I'll try to think something else.

2022/Q3/artifacts/PoS-tla/MC_Staking.tla
angbrav: Why should I? I am not sure I am following.

2022/Q3/artifacts/PoS-tla/MC_Staking.tla
angbrav: Well, I could not come up with something better. We could do something like "IF there is a slash AND slash.finalRate > 0 THEN slashPool > 0", but not sure how useful is that.

PR #30, opened 2022-11-29
draft redelegation pseudocode

tzemanovic:
regarding https://github.com/anoma/namada/issues/34

- Add a new delegator action `redelegate(delegator: Address, src_validator: Address, dest_validator: Address)` that allows to fully re-delegate a delegation bond (partial re-delegation is not allowed):
	- find the current amount of the bond - i.e. apply rewards and processed slashes to the `src_validator` delegation, if any (for delegations, this normally happens only when withdrawing to avoid iteration)
	- at pipeline offset, deduct re-delegated bond's amount from `src_validator`'s total stake and add it to the `dest_validator`
	- for the source delegation's `src_validator`, record re-delegation with the `dest_validator`, the full amount of tokens that's being re-delegated and start and end epochs of the source delegation (the end epoch would be at the pipeline offset - 1)
		-  Note that a validator can have many re-delegations, any slashes to source validator has to iterate all of these when being processed, but we can apply the slash to each destination validator once.
	- For the target delegation, record the `src_validator` and the start and end epoch of the source delegation
		- A bond can only have one re-delegation at the time - while there's a re-delegation on a bond with `end epoch + unbonding_len > current epoch` , it cannot be re-delegated again
		- After that, the bond can be re-delegated, but additionally it has to:
			- Check if there are any rewards and slashes for the re-delegation's source `src_validator` while the re-delegation's amount still contributed to the source validator's stake, apply them together with the re-delegation's target `dest_validator` rewards and slashes
			- The record of the old re-delegation should then be deleted in both the previous re-delegation's source `src_validator`'s records and the target `dest_validator`
	- On a new slashable event from the `src_validator` that has occurred somewhere between the re-delegation's source start and epoch epochs, once the slash is processed in the protocol, it will apply the slash rate on the amount of the re-delegation's target `dest_validator`'s total stake
	- A re-delegation must also be checked when a delegation is being withdrawn using the re-delegation record on the bond, if any, to find any slashes applicable to the `src_validator`
- Validator self-bond naturally cannot be re-delegated as validators are not allowed to delegate

### DIFF
--- a/2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
+++ b/2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
@@ -214,8 +214,8 @@ tx_redelegate(src_validator_address, dest_validator_address, delegator_address)
     // Check that `incoming_redelegations[delegator_address]` for `dest_validator_address` and 
     // `outgoing_redelegations[delegator_address]` for `src_validator_address` either don't exist
     // or if they do, they cannot be slashed anymore (`end + unbonding_length <= cur_epoch`)
-    var incoming = validators[dest_validator_address].incoming_redelegations[delegator_address]
-    var outgoing = validators[src_validator_address].outgoing_redelegations[delegator_address]
+    var incoming = validators[dest_validator_address].incoming_redelegations[src_validator_address][delegator_address]
+    var outgoing = validators[src_validator_address].outgoing_redelegations[dest_validator_address][delegator_address]
     if ((incoming != ⊥ && incoming.end + unbonding_length > cur_epoch) ||
        (outgoing != ⊥ && outgoing.end + unbonding_length > cur_epoch)) then
       return
@@ -238,11 +238,11 @@ tx_redelegate(src_validator_address, dest_validator_address, delegator_address)
     var start = cur_epoch
     var end = cur_epoch + unbonding_length
     // Add a new record in `outgoing_redelegations` for `src_validator_address` 
-    validators[src_validator_address].outgoing_redelegations[delegator_address] =
+    validators[src_validator_address].outgoing_redelegations[dest_validator_address][delegator_address] =
       Redelegation{ amount: bonded_tokens, start, end}
 
     // Add a new record in `incoming_redelegations` for `dest_validator_address` 
-    validators[dest_validator_address].incoming_redelegations[delegator_address] =
+    validators[dest_validator_address].incoming_redelegations[src_validator_address][delegator_address] =
       Redelegation{ amount: bonded_tokens, start, end}
 
     // Add a bond in the `dest_validator_address` for the amount
@@ -352,6 +352,17 @@ func withdraw(validator_address, delegator_address)
         updated_amount -= slashed_amount.amount
         computed_amounts = computed_amounts \ {slashed_amount}
       computed_amounts = computed_amounts \union {SlashedAmount{epoch: slash.epoch, amount: updated_amount*slash.rate}}
+
+      // Regard validator's incoming_redelegations - there might be slashes on 
+      // the source validator that should be applied to this bond
+      forall (<dest, redelegations> in validator[validator_address].incoming_redelegations) do
+        var total_redelegated_and_slashable = 0
+        forall (<owner, redelegation> in redelegations | owner == delegator_address) do
+        // Find any redelegation where slash.epoch is between their start and end epoch
+        if (redelegation.start <= slash.epoch && slash.epoch <= redelegation.end) then
+          total_redelegated_and_slashable += redelegation.amount
+        // Add the slash from redelegation source validator to `computed_amounts`
+        computed_amounts = computed_amounts \union {SlashedAmount{epoch: slash.epoch, amount: total_redelegated_and_slashable*slash.rate}}
 
     var amount_after_slashing = updated_amount - sum({computed_amount.amount | computed_amount in computed_amounts})
     balance[delegator_address] += amount_after_slashing
@@ -544,7 +555,8 @@ end_of_epoch()
       append(slashes[validator_address], slash)
       var total_staked = read_epoched_field(validators[validator_address].total_deltas, slash.epoch, 0)
 
-      // Regard outgoing_redelegations
+      // Regard outgoing_redelegations - the slashed amount for these is applied 
+      // on the destination validator
       forall (<dest, redelegations> in validator[validator_address].outgoing_redelegations) do
         var total_redelegated_and_slashable = 0
         forall (<owner, redelegation> in redelegations) do
--- a/2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
+++ b/2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
@@ -354,10 +354,10 @@ func withdraw(validator_address, delegator_address)
     var total_redelegated_and_slashable = 0
     forall (<src, redelegations> in validators[validator_address].incoming_redelegations) do
       forall (<owner, redelegation> in redelegations | owner == delegator_address) do
-        forall (<start> bonds[owner][src].deltas)
+        forall (<start, end, bond_amount> in bonds[owner][src].deltas)
           // Find any redelegation where slash.epoch is between the src bond start and redelegation start epoch
           forall (src_slash in slashes[src] in slash.epoch order s.t start <= slash.epoch && slash.epoch < redelegation.end)
-            total_redelegated_and_slashable += redelegation.amount
+            total_redelegated_and_slashable += bond_amount
             // Add the slash from redelegation source validator to `computed_amounts`
             computed_amounts = computed_amounts \union {SlashedAmount{epoch: slash.epoch, amount: total_redelegated_and_slashable*slash.rate}}
 
--- a/2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
+++ b/2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
@@ -26,8 +26,6 @@ type Validator struct {
   reward_address Addr
   jail_record JailRecord
   frozen map<Epoch, bool>
-  outgoing_redelegations map<dest:Addr, map<owner:Addr, Redelgation>>
-  incoming_redelegations map<src:Addr, map<owner:Addr, Redelgation>>
 }
 
 type Bond struct {
@@ -56,12 +54,6 @@ type WeightedValidator struct {
 type ValidatorSet struct {
   active orderedset<WeightedValidator>
   inactive orderedset<WeightedValidator>
-}
-
-type Redelgation struct {
-  amount int
-  start Epoch
-  end Epoch
 }
 ```
 
@@ -160,14 +152,14 @@ tx_unjail(validator_address)
 ```go
 tx_self_bond(validator_address, amount)
 {
-  bond(validator_address, validator_address, amount)
+  bond(validator_address, validator_address, amount, pipeline_length)
 }
 ```
 
 ```go
 tx_unbond(validator_address, amount)
 {
-  unbond(validator_address, validator_address, amount, true)
+  unbond(validator_address, validator_address, amount)
 ```
 
 ```go
@@ -196,55 +188,19 @@ tx_delegate(validator_address, delegator_address, amount)
 ```go
 tx_undelegate(validator_address, delegator_address, amount)
 {
-  unbond(validator_address, delegator_address, amount, true)
+  unbond(validator_address, delegator_address, amount)
 }
 ```
 
+<!-- 
 ```go
-// Amount is not specified - a re-delegation must transfer the full amount
-tx_redelegate(src_validator_address, dest_validator_address, delegator_address)
+tx_redelegate(src_validator_address, dest_validator_address, delegator_address, amount)
 {
-
-  // Disallow re-delegation if either validator is frozen (similar to unbonding)
-  var src_frozen = read_epoched_field(validators[src_validator_address].frozen, cur_epoch, false)
-  var dest_frozen = read_epoched_field(validators[dest_validator_address].frozen, cur_epoch, false)
-  if (is_validator(src_validator_address, cur_epoch+pipeline_length) && src_frozen == false &&
-      is_validator(dest_validator_address, cur_epoch+pipeline_length) && dest_frozen == false
-  ) then
-    // Check that `incoming_redelegations[delegator_address]` for `dest_validator_address` and 
-    // `outgoing_redelegations[delegator_address]` for `src_validator_address` either don't exist
-    // or if they do, they cannot be slashed anymore (`end + unbonding_length <= cur_epoch`)
-    var incoming = validators[dest_validator_address].incoming_redelegations[src_validator_address][delegator_address]
-    var outgoing = validators[src_validator_address].outgoing_redelegations[dest_validator_address][delegator_address]
-    if ((incoming != ⊥ && incoming.end + unbonding_length > cur_epoch) ||
-       (outgoing != ⊥ && outgoing.end + unbonding_length > cur_epoch)) then
-      return
-
-    // Find the sum of bonded tokens to `src_validator_address`
-    var bonded_tokens = 0
-    var delbonds = {<start, amount> | amount = bonds[delegator_address][src_validator_address].deltas[(start, ⊥)] > 0 && start <= cur_epoch + unbonding_length}
-    forall (<start,amount> in delbonds) do
-      bonded_tokens += amount
-
-    // Unbond the tokens from `src_validator_address` at pipeline offset, but 
-    // without recording it in the `unbonds[delegator_address][src_validator_address]`
-    var record_unbonds = false
-    unbond(src_validator_address, delegator_address, bonded_tokens, record_unbonds)
-
-    var start = cur_epoch
-    var end = cur_epoch + pipeline_length
-    // Add a new record in `outgoing_redelegations` for `src_validator_address` 
-    validators[src_validator_address].outgoing_redelegations[dest_validator_address][delegator_address] =
-      Redelegation{ amount: bonded_tokens, start, end}
-
-    // Add a new record in `incoming_redelegations` for `dest_validator_address` 
-    validators[dest_validator_address].incoming_redelegations[src_validator_address][delegator_address] =
-      Redelegation{ amount: bonded_tokens, start, end}
-
-    // Add a bond in the `dest_validator_address` for the amount
-    bond(dest_validator_address, delegator_address, bonded_tokens)
+  unbond(src_validator_address, delegator_address, amount)
+  bond(dst_validator_address, delegator_address, amount, unbonding_length)
 }
 ```
+-->
 
 ```go
 tx_withdraw_unbonds_delegator(delegator_address)
@@ -293,7 +249,7 @@ func bond(validator_address, delegator_address, amount)
     conclusion: it is an actual issue, unresolved
 */
 //This function is called by transactions tx_unbond, tx_undelegate and tx_redelegate
-func unbond(validator_address, delegator_address, unbond_amount, record_unbonds)
+func unbond(validator_address, delegator_address, unbond_amount)
 {
   //disallow unbonding if the validator is frozen
   var frozen = read_epoched_field(validators[validator_address].frozen, cur_epoch, false)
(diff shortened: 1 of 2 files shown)


2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -188,19 +196,59 @@ tx_delegate(validator_address, delegator_address, amount)
...
+
+    // Unbond the tokens from `src_validator_address` at pipeline offset, but 
+    // without recording it in the `unbonds[delegator_address][src_validator_address]`
+    var record_unbonds = false
+    unbond(src_validator_address, delegator_address, bonded_tokens, record_unbonds)
+
+    var start = cur_epoch
+    var end = cur_epoch + unbonding_length
+    // Add a new record in `outgoing_redelegations` for `src_validator_address` 
+    validators[src_validator_address].outgoing_redelegations[dest_validator_address][delegator_address] =
```
jmalicevic: Why does the source validator need to keep track of where the redelegation goes? Is this used afterwards for something?

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -188,19 +196,59 @@ tx_delegate(validator_address, delegator_address, amount)
...
+      // Apply slashes (and rewards - but we're not dealing with these here)
+      var amount_after_slashing = amount
+      forall (slash in slashes[validator_address] s.t. start <= slash.epoch)
+        amount_after_slashing -= amount*slash.rate
+      bonded_tokens += amount_after_slashing
+
+    // Unbond the tokens from `src_validator_address` at pipeline offset, but 
+    // without recording it in the `unbonds[delegator_address][src_validator_address]`
+    var record_unbonds = false
+    unbond(src_validator_address, delegator_address, bonded_tokens, record_unbonds)
```
jmalicevic: I might be wrong here but is the `unbond` function not removing the slashed amount from the `bonded_tokens` parameter? Will this lead to this being removed twice?

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
tzemanovic: yup, we're using these in `end_of_epoch` for slashing logic to "transfer" slashes for re-delegations that contributed to a slashable event on the source validator, but are now bonded in the destination validator

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
tzemanovic: Good question! I think we're applying slashes on the amount that's being used to update for `update_total_deltas`, but not on the `unbonds` record (this is only set when the newly added `record_unbonds` bool here is true - from this call it's false). This is because the validator total deltas are slashed as soon as the slash is processed, but bonds are only slashed once withdrawn

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -277,7 +325,8 @@ func unbond(validator_address, delegator_address, unbond_amount)
           remain -= amount
           forall (slash in slashes[validator_address] s.t. start <= slash.epoch)
             amount_after_slashing -= amount*slash.rate
-      unbonds[delegator_address][validator_address].deltas[(start,cur_epoch+pipeline_length+unbonding_length)] += unbond_amount
+      if record_unbonds then
+        unbonds[delegator_address][validator_address].deltas[(start,cur_epoch+pipeline_length+unbonding_length)] += unbond_amount
```
angbrav: There was a typo that https://github.com/informalsystems/partnership-heliax/pull/32 fixes. I understand the logic though. Basically, lines 317 and 324 should be unbonds and not bonds, so those should be the ones guarded with record_unbonds. Line 329 is a typo and should not be there.

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
angbrav: I think that the amount to be unbonded should be the amount before slashing. The unbond function takes care of slashing if required.

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -188,19 +196,59 @@ tx_delegate(validator_address, delegator_address, amount)
...
+    // Add a new record in `outgoing_redelegations` for `src_validator_address` 
+    validators[src_validator_address].outgoing_redelegations[dest_validator_address][delegator_address] =
+      Redelegation{ amount: bonded_tokens, start, end}
+
+    // Add a new record in `incoming_redelegations` for `dest_validator_address` 
+    validators[dest_validator_address].incoming_redelegations[src_validator_address][delegator_address] =
+      Redelegation{ amount: bonded_tokens, start, end}
+
+    // Add a bond in the `dest_validator_address` for the amount
+    bond(dest_validator_address, delegator_address, bonded_tokens, pipeline_length)
```
angbrav: The function bond has three arguments: the offset argument does not exist anymore. It uses pipeline_length by default.

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -495,19 +555,21 @@ end_of_epoch()
...
+          // Find any redelegation where slash.epoch is between their start and end epoch
+          if (redelegation.start <= slash.epoch && slash.epoch <= redelegation.end) then
+            total_redelegated_and_slashable += redelegation.amount
+        // Apply the slash on the redelegations' total amount to the `dest` validator
+        apply_slash(dest, slash.rate, total_redelegated_and_slashable)
+        // And subtract it from this validator's `total_staked` amount
+        total_staked -= total_redelegated_and_slashable
+
+      // Apply the slash on this validator
+      apply_slash(validator_address, slash.rate, total_redelegated_and_slashable)
```
angbrav: should not be `total_staked` instead of `total_redelegated_and_slashable`? In my opinion, when slashing the the misbehaving validator, the logic should be exactly the same than before: `total_unbonded` accounts already for what's been unbonded (including redelegations).

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -518,6 +580,25 @@ end_of_epoch()
...
+{
+  var total_unbonded = 0
+  //find the total unbonded from the slash epoch up to the current epoch first
+  forall (epoch in slash.epoch+1..cur_epoch+1) do
+    total_unbonded += validators[validator_address].total_unbonded[epoch]
+
+  var last_slash = 0
+  forall (offset in 1..unbonding_length) do
+    total_unbonded += validators[validator_address].total_unbonded[cur_epoch + offset]
+    var this_slash = (total_staked - total_unbonded) * rate
```
angbrav: Do you mean `staked_amount` instead of `total_staked`? Otherwise I believe that  `total_staked` is undefined.

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -495,19 +555,21 @@ end_of_epoch()
...
+      // Regard outgoing_redelegations - the slashed amount for these is applied 
+      // on the destination validator
+      forall (<dest, redelegations> in validator[validator_address].outgoing_redelegations) do
+        var total_redelegated_and_slashable = 0
+        forall (<owner, redelegation> in redelegations) do
+          // Find any redelegation where slash.epoch is between their start and end epoch
+          if (redelegation.start <= slash.epoch && slash.epoch <= redelegation.end) then
+            total_redelegated_and_slashable += redelegation.amount
+        // Apply the slash on the redelegations' total amount to the `dest` validator
+        apply_slash(dest, slash.rate, total_redelegated_and_slashable)
```
angbrav: I am quite confused by this line. I think it is tricky to get it right. The `apply_slash` function is meant to slash the total stake of a validator at a given epoch as a whole. What we need to do here, conceptually, is to slash a portion of the destination validator: the portion determined by the redelegation. I am not sure `apply_slash` does that, e.g., it subtracts total_unbonded which is a sum across all users. Also, how does this deal with a chain of redelegations? (see my comment above)

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -188,19 +196,59 @@ tx_delegate(validator_address, delegator_address, amount)
...
+
+  // Disallow re-delegation if either validator is frozen (similar to unbonding)
+  var src_frozen = read_epoched_field(validators[src_validator_address].frozen, cur_epoch, false)
+  var dest_frozen = read_epoched_field(validators[dest_validator_address].frozen, cur_epoch, false)
+  if (is_validator(src_validator_address, cur_epoch+pipeline_length) && src_frozen == false &&
+      is_validator(dest_validator_address, cur_epoch+pipeline_length) && dest_frozen == false
+  ) then
+    // Check that `incoming_redelegations[delegator_address]` for `dest_validator_address` and 
+    // `outgoing_redelegations[delegator_address]` for `src_validator_address` either don't exist
+    // or if they do, they cannot be slashed anymore (`end + unbonding_length <= cur_epoch`)
```
angbrav: To clarify. This guarantees that no evidence will be submitted affecting the previous redelegation. Let me do some math. The `src_validator` can misbehave with the redelegated tokens up to an epoch  `e < cur_epoch+pipeline_length` (at that epoch the tokens already do not contribute to its voting power). The evidence for that infraction should be submitted by `e+unbonding_length` at the latest. Is that correct? In that case, shouldn't the end of a redelegation set to `cur_epoch+pipeline_length` instead of `cur_epoch+unbonding_length`?

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -188,19 +196,59 @@ tx_delegate(validator_address, delegator_address, amount)
...
+        amount_after_slashing -= amount*slash.rate
+      bonded_tokens += amount_after_slashing
+
+    // Unbond the tokens from `src_validator_address` at pipeline offset, but 
+    // without recording it in the `unbonds[delegator_address][src_validator_address]`
+    var record_unbonds = false
+    unbond(src_validator_address, delegator_address, bonded_tokens, record_unbonds)
+
+    var start = cur_epoch
+    var end = cur_epoch + unbonding_length
```
angbrav: Why `end = cur_epoch + unbonding_length` and not `end = cur_epoch + pipeline_length`?. The redelegation is completed at `cur_epoch + pipeline_length`

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
angbrav: Also, this does not prevent having chained redelegations. For instance, a user could do redelegation from `val1` to `val2` at `e`. Then at `e+pipeline_length` from `val2` to `val3`. I feel that for some values of `pipeline_length` and `unbonding_length` this could be a problem. Did I misunderstand something here? is this an issue? Maybe offsets prevent this situation, but it does not look like.

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -304,6 +353,17 @@ func withdraw(validator_address, delegator_address)
         computed_amounts = computed_amounts \ {slashed_amount}
       computed_amounts = computed_amounts \union {SlashedAmount{epoch: slash.epoch, amount: updated_amount*slash.rate}}
 
+      // Regard validator's incoming_redelegations - there might be slashes on 
+      // the source validator that should be applied to this bond
+      forall (<dest, redelegations> in validator[validator_address].incoming_redelegations) do
+        var total_redelegated_and_slashable = 0
+        forall (<owner, redelegation> in redelegations | owner == delegator_address) do
+        // Find any redelegation where slash.epoch is between their start and end epoch
+        if (redelegation.start <= slash.epoch && slash.epoch <= redelegation.end) then
```
angbrav: I wonder if this is correct. What if the `src_validator` misbehaved before `redelegation.start` but the evidence was found after `redelegation.start`? I guess this is possible and we should slash.

angbrav (changes requested):
Thanks @tzemanovic. I have a lot of comments and I am not sure I fully understand everything - it is quite complex! I think it would be great to start by clarifying this first round of questions/comments and then do a second iteration.

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
tzemanovic: I'll rebase this PR, thanks! I'll remove the line 329 and put the condition `if record_unbonds...` onto the two unbonds updates on 317 and 324

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
tzemanovic: good point! will fix it

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
tzemanovic: ah, I think I copied it from `tx_self_bond` - I'll fix it there too

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
tzemanovic: in `tx_redelegate`, the call to unbond uses:
```
    var record_unbonds = false
    unbond(src_validator_address, delegator_address, bonded_tokens, record_unbonds)
```
because re-delegations won't be withdrawable, so we don't record them as unbonds. I think those would not be reflected in `total_unbonded`, but not sure if the logic is quite right

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
tzemanovic: ah, yes!

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
tzemanovic: I see, yeah it's most likely not right then. I agree with the description of what we need to do. 

Rwgarding a chain of re-delegations, we prohibit re-re-delegation while a re-delegated bond is slashable from the re-delegation source (and we also require that re-delegation must transfer a bond in full). After that, it should be no longer relevant so should be ok

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
tzemanovic: Ah, yes, the re-delegation end should indeed be set to `cur_epoch+pipeline`.

> For instance, a user could do redelegation from val1 to val2 at e. Then at e+pipeline_length from val2 to val3.

I think if it's re-delegated at `e`, then the earliest epoch for the next allowed re-delegation should be `e + pipeline + unbonding`

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
tzemanovic: yup, this should be pipeline

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
tzemanovic: Ah, I think for the `incoming_redelegations`, we have to check slashes on the redelegation `src` address and like you say, the epoch range check should be different - I think this may also be a bit tricky as a re-delegation may applied on a bond with multiple `start` epochs.

There's also a typo on the unchanged line:
```
    forall (slash in slashes in slash.epoch order s.t. start <= slash.epoch && slash.epoch <= end) do
```
should be
```
    forall (slash in slashes[validator_address] in slash.epoch order s.t. start <= slash.epoch && slash.epoch <= end) do
```

tzemanovic: I've pushed a different version that attempts to check slashes and use the bonds on the redelegation source to determine slashed amount on re-delegation target. This however doesn't work as is, because we're removing the bonds when we re-delegate, so that information is gone (we changed this couple times before), but we might actually need to preserve it for this

angbrav:
@tzemanovic I have moved your changes to the model in the Q4 folder. You were working on the Q3 folder, that's probably why the model was not completely up-to-date. Sorry, it is a complicated way of organizing things :S

Unfortunately, now the comments are not connected to any specific line, but I couldn't figure out how to move the changes without messing that up.

angbrav:
@tzemanovic I have been working on this and I am starting to think that this is quite complicated. I have two main concerns/comments (let me know if something is unclear):

1. I do not understand the logic in withdraw. I think we need to keep more state to know which unbonds are coming from a redelegation. Right now, it seems that you slash everything that was redelegated, but the user may not have unbonded those tokens. Note that this is not solved by adding the start epoch of each redelegated bond, which needs to be done to slash validators correctly at the end of an epoch. For instance:
    - Assume that a user redelegates `X` tokens from `val1` to `val2`.
    - At some epoch later, assume that the user unbonds `Y` tokens from `val2` and assume that `X<Y`.
    -  Assume that `val1` misbehaved while holding those `X` tokens.
    -  Assume that the user withdraws the `Y` unbonded tokens.
    -  Isn't the user going to be slashed the `X` tokens?  This would over-slash the user.
    -     -     - 
2. Second, I still believe that there might be a problem with chain redelegations: I do not think it is prevented. Let me elaborate a bit. Assume that a user has X tokens delegated to a validator `val1` by epoch `e`. Assume that the user redelegates this to a validator `val2` by `e` and then redelegates them to a validator `val3` by epoch `e+1`. I think the following code does not disallow such an execution:
```go
var incoming = validators[dest_validator_address].incoming_redelegations[src_validator_address][delegator_address]
    var outgoing = validators[src_validator_address].outgoing_redelegations[dest_validator_address][delegator_address]
    if ((incoming != ⊥ && incoming.end + unbonding_length > cur_epoch) ||
       (outgoing != ⊥ && outgoing.end + unbonding_length > cur_epoch)) then
      return
```
My concern is that if`val1` misbehaves, we need to slash `val3`. Nevertheless, this won't happen: we will only try to slash `val2`. Am I missing something? I think chained redelegations would be problem also in withdraw, so we should find a way of disallow them if possible.

- - -

I am not sure what's the solution for either of these issues but it seems quite complicated. I was wondering how important is for you to be able to pipeline redelegation. I am under the impression that things could get much easier if the redelegated tokens are only bonded to the dest validator after the unbonding period (or perhaps pipeline+unbonding). My intuition is that
this way, one does not need to carry over slashes across redelegations. One option would be instead of creating the new bonds immediately, one could enqueue the redelagation until any infraction could have been found for the redelegated tokens. Then, at the end of the last epoch at which an infraction may be processed, we create the bonds in the destination validator after slashing any tokens due to infractions of the source validator. This is just an idea that I believe could simplify things quite a bit.

2022/Q3/artifacts/PoS-pseudocode/PoS-model.md
angbrav: I agree with this (see https://github.com/informalsystems/partnership-heliax/pull/30#issuecomment-1409410637). The model does not disallow this currently, right?

tzemanovic: I think that should be handled by this: https://github.com/informalsystems/partnership-heliax/pull/30/files#diff-60fc868044555c2c663475a5ee05bddd58e4984de04344c8b1cad25fac2fd935R214-R221

tzemanovic: ah nvm, let me respond to the comment below

tzemanovic:
@angbrav Agree that this is quite complicated to fit in with the design!

1. Ah yes, I agree that the re-delegation withdrawal slashing is not right, but to clarify - in the example, it is the `val1` that misbehaved, so when withdrawing `Y` tokens only `X*slash.rate` should be slashed, right?
 
If we keep the data of start epoch of each re-delegated bond delta, couldn't we use it instead of the `bonds[owner][src].deltas` to correctly apply it only over what's actually beeb re-delegated?

Related to this, I think I have a mistake on line 563:

```
apply_slash(validator_address, slash.rate, total_redelegated_and_slashable)
```

should be

```
apply_slash(validator_address, slash.rate, total_staked)
```

That is the validator's stake, minus what's been re-delegated. IIUC, the `apply_slash` function should then regard any unbonds.

2. Ah, yes, the chain of re-delegations must be prevented and the check is not quite right, but I think this should be possible with the data that we have. We should really be checking that the `validators[src_validator_address]` doesn't have any `incoming_redelegations` from `delegator_address` where `incoming.end + unbonding_length > cur_epoch`. For the `dest_validator_address` I think it's okay to allow multiple re-delegations from different source validators into the same `dest_validator_address`, unless I'm still missing something.

It's a good idea to make re-delegations apply at `pipeline + unbonding` and it would definitely simply this, but I'm not sure how feasible that is as we're planning to set the initial `unbonding_len` quite high, so it would then take ~3 weeks to re-delegate. While there is extra tracking and processing required, the pipeline will be only ~2 days, so if the validator you're delegating to suddenly drops out, you can change reasonably fast

angbrav:
> @angbrav Agree that this is quite complicated to fit in with the design!
> 
> 1. Ah yes, I agree that the re-delegation withdrawal slashing is not right, but to clarify - in the example, it is the `val1` that misbehaved, so when withdrawing `Y` tokens only `X*slash.rate` should be slashed, right?

Yes, I was assuming that slash.rate was 1 for simplicity. I should have spelled this out.
 
> Related to this, I think I have a mistake on line 563:
> 
> ```
> apply_slash(validator_address, slash.rate, total_redelegated_and_slashable)
> ```
> 
> should be
> 
> ```
> apply_slash(validator_address, slash.rate, total_staked)
> ```
> 

Yes!

angbrav:
Closed by #45.

PR #32, opened 2022-12-15
Fix inconsistencies psuedocode

angbrav:
It fixes a couple of inconsistencies:
- Wrong range when computing total_unbonded in end_of_epoch
- A couple of typos in unbond
- Fix an offset problem in unbond: unbonds end at cur_epoch+pipeline_length+unbonding_length
- Fix an offset problem when slashing at withdraw: unbond.end of an unbond determine when the user can withdraw but those tokens stop contributing to the validators voting power at unbond.end - unbonding_length


tzemanovic (approved):
LGTM, thank you!

Issue #34, opened 2023-02-02
Validator's slashing is wrong due to ongoing unbonds and bonds


angbrav:
I think I have found an issue with the way we are computing slashing for validators at the end of an epoch. I think the issue is related to the fact that bonds can be unbonded before these are reflected in `totalDeltas`. See the following:

Assume that `pipeline_length=1`and `unbonding_lenght=2`. Assume that `slash_rate=1` for simplicity and a single `user` and a validator `val`.
Assume the following execution:
- `epoch=0`:
  - `user` delegates `10` tokens to `val`.
  - `val.totalDeltas[0]= 0, val.totalDeltas[>=1]=10`
  - `bonds[user][val] = {[amount |-> 10, start |-> 1]}
- `epoch=3`
  -  `user` delegates `12` tokens to `val`.
  - `val.totalDeltas[0]=0, val.totalDeltas[1,2,3]=10, val.totalDeltas[>=4]=22`
  - `bonds[user][val] = {[amount |-> 10, start |-> 1], [amount |-> 12, start |-> 4]}
- `epoch=4`
  -  `user` unbonds `22` tokens from `val`.
  - `val.totalDeltas[0]=0, val.totalDeltas[1,2,3]=10, val.totalDeltas[4]=22`, val.totalDeltas[>=5]=0`
  - `bonds[user][val] = {}`
  - `unbonds[user][val] = {[amount |-> 10, start |-> 1, end |-> 7], [amount |-> 12, start |-> 4,  end |-> 7]}`
  - `val.totalUnbonded[0,1,2,3,4]=0, val.totalUnbonded[5]=22`
  - evidence submitted for `val` at `epoch=3`. Will be processed at the end of epoch 5.
- `epoch=5`
  -  process evidence in `end_of_epoch()` function
  - `total_staked=val.totalDeltas[3]=10`
  - `total_unbonded=val.totalUnbonded[4]+val.totalUnbonded[5]=22`
  - for `offset=1` -> `total_unbonded=total_unbonded+val.totalUnbonded[6]=22`
                                  `this_slash=total_staked-total_unbonded=10-22=-12`
                                  `diff_slashed_amount=last_slash - this_slash=0 - (-12)= 12`
  - `val.totalDeltas[0]=0, val.totalDeltas[1,2,3]=10, val.totalDeltas[4,5]=22, val.totalDeltas[>=6]=12`
                              
 We have to figure how to compute slashes correctly.

angbrav:
[...]

PR #37, opened 2023-02-09
Fix issue 34 

angbrav:
This PR fixes issue #34 in the pseudocode model.

The current model disregards a corner case that should be taken care of in the implementation:
- Assume a user delegates 10 tokens to a validator at epoch `e1`
- Assume the user unbonds twice 5 tokens from the same validator in the same epoch in two different transactions `e2`
- When executing the first undelegate tx, the model creates the record `UnbondRecord{amount: 5, start: e1 }` and updates the bond to 5 tokens
- When executing the second undelegate tx, the model creates the same record `UnbondRecord{amount: 5, start: e1 }` and removes the bond.
- The problem is that we keep unbond records in a set and when we try to add the second record, since it is a duplicate, it will be discarded.

It is an easy fix I'd say: use a bag instead of a set to allow duplicates, or check if the set includes the record and act upon (remove it, create a new one with double the amount, and add it). For now, I've left a comment on the model; I didn't want to "pollute" it. If we decide not to deal with it in the mode, I'll open an issue and refer to it in the model.

### DIFF
--- a/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
+++ b/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
@@ -31,7 +31,7 @@ type Validator struct {
   consensus_key map<Epoch, Key>
   state map<Epoch, {inactive, candidate}>
   total_deltas map<Epoch, amount:int>
-  total_unbonds map<Epoch, amount:int>
+  total_unbonds map<Epoch, Set<UnbondRecord>>
   voting_power map<Epoch, VotingPower>
   reward_address Addr
   jail_record JailRecord
@@ -519,12 +519,12 @@ end_of_epoch()
       //find the total unbonded from the slash epoch up to the current epoch first
       //a..b notation determines an integer range: all integers between a and b inclusive
       forall (epoch in slash.epoch+1..cur_epoch) do
-        forall (unbond in validators[validator_address].total_unbonded[epoch] s.t. unbond.start <= slash.epoch)
+        forall (unbond in validators[validator_address].total_unbonds[epoch] s.t. unbond.start <= slash.epoch)
           total_unbonded += unbond.amount
 
       var last_slash = 0
       forall (offset in 1..unbonding_length) do
-        forall (unbond in validators[validator_address].total_unbonded[epoch] s.t. unbond.start <= slash.epoch)
+        forall (unbond in validators[validator_address].total_unbonds[epoch] s.t. unbond.start <= slash.epoch)
           total_unbonded += unbond.amount
         var this_slash = (total_staked - total_unbonded) * slash.rate
         var diff_slashed_amount = last_slash - this_slash
--- a/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
+++ b/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
@@ -22,6 +22,11 @@ type UnbondRecord struct {
   start Epoch
 }
 
+type SlashedAmount struct {
+  epoch Epoch
+  amount uint
+}
+
 type Validator struct {
   consensus_key map<Epoch, Key>
   state map<Epoch, {inactive, candidate}>
--- a/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
+++ b/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
@@ -513,7 +513,7 @@ end_of_epoch()
       forall (offset in 1..unbonding_length) do
         forall (unbond in validators[validator_address].total_unbonded[epoch] s.t. unbond.start <= slash.epoch)
           total_unbonded += unbond.amount
-        var this_slash = (total_staked + total_bonded - total_unbonded) * slash.rate
+        var this_slash = (total_staked - total_unbonded) * slash.rate
         var diff_slashed_amount = last_slash - this_slash
         last_slash = this_slash
         update_total_deltas(validator_address, offset, diff_slashed_amount)
--- a/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
+++ b/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
@@ -17,12 +17,16 @@ type JailRecord struct {
   epoch Epoch
 }
 
+type UnbondRecord struct {
+  amount uint
+  start Epoch
+}
+
 type Validator struct {
   consensus_key map<Epoch, Key>
   state map<Epoch, {inactive, candidate}>
   total_deltas map<Epoch, amount:int>
   total_unbonds map<Epoch, amount:int>
-  total_bonds map<Epoch, amount:int>
   voting_power map<Epoch, VotingPower>
   reward_address Addr
   jail_record JailRecord
@@ -231,7 +235,6 @@ func bond(validator_address, delegator_address, amount)
     //debit amount from delegator account and credit it to the PoS account
     balances[delegator_address] -= amount
     balances[pos] += amount
-    validators[validator_address].total_bonds[cur_epoch+pipeline_length] += amount
     update_total_deltas(validator_address, pipeline_lenght, amount)
     update_voting_power(validator_address, pipeline_lenght)
     update_total_voting_power(pipeline_lenght)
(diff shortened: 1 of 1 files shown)


2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -497,15 +503,18 @@ end_of_epoch()
       var total_staked = read_epoched_field(validators[validator_address].total_deltas, slash.epoch, 0)
 
       var total_unbonded = 0
+      var total_bonded = 0
```
angbrav:
```suggestion
```

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -500,12 +506,14 @@ end_of_epoch()
...
+        forall (unbond in validators[validator_address].total_unbonded[epoch] s.t. unbond.start <= slash.epoch)
+          total_unbonded += unbond.amount
 
       var last_slash = 0
       forall (offset in 1..unbonding_length) do
-        total_unbonded += validators[validator_address].total_unbonded[cur_epoch + offset]
-        var this_slash = (total_staked - total_unbonded) * slash.rate
+        forall (unbond in validators[validator_address].total_unbonded[epoch] s.t. unbond.start <= slash.epoch)
+          total_unbonded += unbond.amount
+        var this_slash = (total_staked + total_bonded - total_unbonded) * slash.rate
```
angbrav:
```suggestion
        var this_slash = (total_staked - total_unbonded) * slash.rate
```

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -272,14 +277,23 @@ func unbond(validator_address, delegator_address, unbond_amount)
           remain = 0
```
brentstone: Should `remain = 0` come after the loop over the slashes and `validators[validator_address].total_unbonds[cur_epoch+pipeline_length] = {UnbondRecord{amount: remain, start: start}}`?

angbrav: Thanks @brentstone! You are completely right. Fixed.

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -17,6 +17,11 @@ type JailRecord struct {
   epoch Epoch
 }
 
+type UnbondRecord struct {
+  amount uint
+  start Epoch
+}
+
 type Validator struct {
```
brentstone, line 30: `total_unbonds` should really be `total_unbonded`, correct?

brentstone:
Going through the pseudocode again as I work on https://github.com/anoma/namada/pull/892, apologies if some of my questions are technically out of scope for this PR but I figure here is a good place to put them anyway...

Looks like in withdrawing we are using a struct that looks like
```
type SlashedAmount struct {
  epoch Epoch
  amount uint
}
```
If I understand this correctly, we should add it to the other type declarations at the top of `PoS-model.md`

brentstone:
What is the meaning of `computed_amounts = computed_amounts \ {slashed_amount}` in `func withdraw`?

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
angbrav: Do you mean the field in the Validator's struct? If I am not wrong, we use `total_unbonds` throughout the spec and `total_unbonded` at the `end-of-epoch` function when folding `total_unbonds`. I agree we could find better names. Was that your concern, or I misunderstood it?

angbrav:
> Going through the pseudocode again as I work on [anoma/namada#892](https://github.com/anoma/namada/pull/892), apologies if some of my questions are technically out of scope for this PR but I figure here is a good place to put them anyway...
> 

Sure, no problem.

> Looks like in withdrawing we are using a struct that looks like
> 
> ```
> type SlashedAmount struct {
>   epoch Epoch
>   amount uint
> }
> ```
> 
> If I understand this correctly, we should add it to the other type declarations at the top of `PoS-model.md`

Yes, done.

> What is the meaning of `computed_amounts = computed_amounts \ {slashed_amount}` in `func withdraw`?

Set difference: X\Y is the set of elements that belong to X but not to Y. In this case it removes `slashed_amount` from the set `computed_amounts`.

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -269,17 +279,26 @@ func unbond(validator_address, delegator_address, unbond_amount)
...
         else if amount <= remain && remain > 0 do
           bonds[delegator_address][validator_address].deltas[start] = 0
           unbonds[delegator_address][validator_address].deltas[start, cur_epoch+pipeline_length+unbonding_length] = amount
-          remain -= amount
+          validators[validator_address].total_unbonds[cur_epoch+pipeline_length] = {UnbondRecord{amount: amount, start: start}} \union validators[validator_address].total_unbonds[cur_epoch+pipeline_length]
           forall (slash in slashes[validator_address] s.t. start <= slash.epoch)
             amount_after_slashing -= amount*slash.rate
-      validators[validator_address].total_unbonds[cur_epoch+pipeline_length] += unbond_amount
+          remain -= amount
       update_total_deltas(validator_address, pipeline_length, -1*amount_after_slashing)
```
brentstone, line 302: I think that we should be tracking something like `total_slashed_amount` rather than `amount_after_slashing`. Looks like we want to supply `update_total_deltas` with the deltas change, rather than the new value post-slashing.

So instead of `amount_after_slashing -= amount*slash.rate`, want `total_slashed_amount += amount*slash.rate`.

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
brentstone: Ok thanks for clarifying that these are intended to be different.

I'll elaborate a bit. First, yes, the field `total_unbonds` inside the `Validator` struct should be a `map<Epoch, UnbondRecord>` judging by the `unbond` code. It also seems that the `Validator` struct should explicitly have a `total_unbonded` field that is a `map<Epoch, uint>`, right? I think I was confused by the lack of a declaration of `total_unbonded` anywhere outside the `end_of_epoch()` code.

I'm still somewhat confused by what the purpose of `Validator::total_unbonds` is. From the pseudocode, it seems only to be used in `func unbond`, where we set values inside of it, but I don't see any other place where its data is read and then used for something. It is not in `end_of_epoch`, but perhaps there is another bug. From the following code from `end_of_epoch`:

```
      var total_unbonded = 0
      //find the total unbonded from the slash epoch up to the current epoch first
      //a..b notation determines an integer range: all integers between a and b inclusive
      forall (epoch in slash.epoch+1..cur_epoch) do
        forall (unbond in validators[validator_address].total_unbonded[epoch] s.t. unbond.start <= slash.epoch)
          total_unbonded += unbond.amount
```
it looks like perhaps `forall (unbond in validators[validator_address].total_unbonded[epoch]` should be `forall (unbond in validators[validator_address].total_unbonds[epoch]`. Does this seem to be the intention?

brentstone: actually I think I'm realizing that `total_unbonds` should be a `map<Epoch, Set<UnbondRecord>>`, where `Set<UnbondRecord>` could also be a `Vec` or similar

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
brentstone: Sorry, realized I am wrong in the above comment.

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
angbrav: Yes, the code in `end_of_epoch` has a typo and validator's `total_unbonds` should be `map<Epoch, Set<UnbondRecord>>`. Both are now fixed, thanks for spotting them! It is difficult to produce typo-free pseudocode for non-executable specs...

PR #38, opened 2023-02-17
New slashing for validator's at the end of an epoch (fix issue #34)

angbrav:
This PR aims at closing #34 by proposing the following changes:
- It renames the validator's field `total_unbonds` to `set_unbonds` because it now stores a set un `UnbondRecord`.
- It defines the function `compute_amount_after_slashing` that takes an unbond amount and a set of slashes and computes the amount after applying the slashes. The function is used when unbonding, withdrawing and at the end of epochs.
- The logic for applying slashes when a user unbonds tokens now matches the logic on withdrawing. Please check.
- The slashing logic at the end of an epoch is new (major change). It now considers that the bonds contributing to the validator when this misbehaved may not have been contributing in its entirety at the misbehaving epoch (they could have been slashed already). The new logic takes this into consideration by:
  - (as before) computing `total_unboded` at an epoch `e` as the sum of all unbonds contributing to the validator stake at the misbehaving epoch that no longer contribute at `e`.
  - (new) applying to each relevant unbond, the set of slashes that could have been applied before the misbehaving epoch.

### DIFF
--- a/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
+++ b/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
@@ -525,7 +525,8 @@ end_of_epoch()
 
       var total_unbonded = 0
       var last_slash = 0
-      forall (offset in 1..unbonding_length) do
+      // up to pipeline_length because there cannot be any unbond in a greater ß (cur_epoch+pipeline_length is the upper bound)
+      forall (offset in 1..pipeline_length) do
         forall (unbond in validators[validator_address].set_unbonds[cur_epoch + offset] s.t. unbond.start <= slash.epoch) do
           union_unbonded = union_unbonded \union {unbond}
         forall (unbond in union_unbonded) do
--- a/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
+++ b/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
@@ -538,7 +538,7 @@ end_of_epoch()
           total_unbonded += compute_amount_after_slashing(set_prev_slashes, unbond.amount)
         var this_slash = (total_staked - total_unbonded) * slash.rate
         var diff_slashed_amount = last_slash - this_slash
-        update_total_deltas(validator_address, offset, slash_amount)
+        update_total_deltas(validator_address, offset, diff_slashed_amount)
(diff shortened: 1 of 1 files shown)


brentstone:
@tzemanovic brought up to me an issue we see in several places (`func unbond`, `func withdraw`, `end_of_epoch`) in the pseudocode that relate to cubic slashing. Particularly, this concerns the `set_slashes` that is collected and then supplied to `compute_amount_after_slashing(set_slashes, amount)`.

We need to consider slashes currently being filtered out from `set_slashes` that may be contributing the to slash rate for a given slash that is inside of `set_slashes` due to cubic slashing. For example, in the `unbond` case, a slash for which `slash.epoch = start - 1` may contribute to the rate if there also exists a slash for which `slash.epoch = start`.

To account for this, we may need to iterate thru the `slashes[validator_address]` twice in order to collect the necessary information.

angbrav:
> @tzemanovic brought up to me an issue we see in several places (`func unbond`, `func withdraw`, `end_of_epoch`) in the pseudocode that relate to cubic slashing. Particularly, this concerns the `set_slashes` that is collected and then supplied to `compute_amount_after_slashing(set_slashes, amount)`.
> 
> We need to consider slashes currently being filtered out from `set_slashes` that may be contributing the to slash rate for a given slash that is inside of `set_slashes` due to cubic slashing. For example, in the `unbond` case, a slash for which `slash.epoch = start - 1` may contribute to the rate if there also exists a slash for which `slash.epoch = start`.
> 
> To account for this, we may need to iterate thru the `slashes[validator_address]` twice in order to collect the necessary information.

I am not sure I am following. The slash rate, as far as we discussed, is computed once when the slash is processed at the end of an epoch. This is done by the two following lines:

```go
//iterate over all slashes for infractions within (-1,+1) epochs range (Step 2.1 of cubic slashing)
var set_slashes = {s | s in enqueued_slashes[epoch] && cur_epoch-1 <= epoch <= cur_epoch+1 && s.validator == validator_address}
//calculate the slash rate (Step 2.2 of cubic slashing)
var rate = compute_final_rate(set_slashes)
```

Has this changed? Am I missing something?

On a separate note, I think this is an issue unrelated to the PR, so it would be better to open a separate issue to discuss it.

PR #42, opened 2023-03-09
PoS pseudocode edits

brentstone:
Just some small adjustments that should be made to the pseudocode.


angbrav (approved):
LGTM! good catch.

PR #43, opened 2023-03-22
More PoS pseudocode edits

brentstone:
Miscellaneous edits and updates to the PoS pseudocode. Also includes updates that align the pseudocode more closely with the Namada implementation. The changes here reflect those that were addressed before discovering the need to adjust procedures in order to prevent validator deltas from going negative. This is in light of the Namada slashing conditions being broader than those allowed for in the TLA+ model. In particular, in the TLA+, a validator can only misbehave once within a single window of `unbonding_len` epochs.

### DIFF
--- a/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
+++ b/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
@@ -271,7 +271,8 @@ func unbond(validator_address, delegator_address, total_amount)
     var delbonds = {<start, amount> | amount = bonds[delegator_address][validator_address].deltas[start] > 0 && start <= cur_epoch + pipeline_len}
     if (sum{amount | <start, amount> in delbonds} >= total_amount) then
       var remain = total_amount
-      var amount_after_slashing = 0
+      var amount_after_slashing = 0 // Track the amount to change the deltas (may be less than `total_amount` due to previous slash processing)
+
       // Iterate over bonds and create unbond
       forall (<start, amount> in delbonds while remain > 0) do
         // Get the minimum of the remainder and the unbond, equal to amount if remain > amount and remain otherwise 
--- a/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
+++ b/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
@@ -278,7 +278,7 @@ func unbond(validator_address, delegator_address, total_amount)
         // Get the minimum of the remainder and the unbond, equal to amount if remain > amount and remain otherwise 
         var amount_unbonded = min{amount, remain}
         bonds[delegator_address][validator_address].deltas[start] = amount - amount_unbonded
-        unbonds[delegator_address][validator_address].deltas[start, cur_epoch+pipeline_length+unbonding_length] = amount_unbonded
+        unbonds[delegator_address][validator_address].deltas[start, cur_epoch+pipeline_length+unbonding_length] += amount_unbonded
         // Set of slashes that happened while the bond was contributing to the validator's stake
         var set_slashes = {s | s in slashes[validator_address] && start <= slash.epoch }
         amount_after_slashing += compute_amount_after_slashing(set_slashes, amount_unbonded)
--- a/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
+++ b/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
@@ -478,7 +478,6 @@ freeze_validator(validator_address)
   var epochs = {epoch | cur_epoch <= epoch <= cur_epoch+unbonding_length &&
                         (epoch > cur_epoch => validators[validator_address].frozen[epoch] != ⊥)}
   forall (epoch in epochs) do
-    var total = read_epoched_field(validators[validator_address].frozen, epoch, false)
     validators[validator_address].frozen[epoch] = true
   //schedule when to unfreeze the validator: after processing the enqueued slash
   validators[validator_address].frozen[cur_epoch+unbonding_length+1] = false
--- a/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
+++ b/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
@@ -523,7 +523,7 @@ end_of_epoch()
       forall (epoch in slash.epoch+1..cur_epoch) do
         forall (unbond in validators[validator_address].set_unbonds[epoch] s.t. unbond.start <= slash.epoch)
           var set_prev_slashes = {s | s in slashes[validator_address] && unbond.start <= s.epoch && s.epoch + unbonding_length < slash.epoch}
-          total_unbonded = compute_amount_after_slashing(set_prev_slashes, unbond.amount)
+          total_unbonded += compute_amount_after_slashing(set_prev_slashes, unbond.amount)
 
       var last_slash = 0
       // up to pipeline_length because there cannot be any unbond in a greater ß (cur_epoch+pipeline_length is the upper bound)
--- a/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
+++ b/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
@@ -98,12 +98,11 @@ total_voting_power[] in Epoch to VotingPower //map from epoch to voting_power
 ## Validator transactions
 
 ```go
-tx_become_validator(validator_address, consensus_key, staking_reward_address)
+tx_become_validator(validator_address, consensus_key)
 {
   // Check that become_validator has not been called before for validator_address
   var state = read_epoched_field(validators[validator_address].state, cur_epoch+pipeline_length, ⊥)
-  if (state == ⊥ && validator_address != staking_reward_address) then
-    validators[validator_address].reward_address = staking_reward_address
+  if (state == ⊥) then
     // Set status to candidate and consensus key at n + pipeline_length
     validators[validator_address].consensus_key[cur_epoch+pipeline_length] = consensus_key
     validators[validator_address].state[cur_epoch+pipeline_length] = candidate
--- a/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
+++ b/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
@@ -55,6 +55,7 @@ type Slash struct {
   validator Addr
   rate float
   voting_power VotingPower // the voting power used to commit the infraction
+  slash_type string
 }
 
 type WeightedValidator struct {
@@ -494,8 +495,8 @@ func is_validator(validator_address, epoch){
 ```go
 func get_min_slash_rate(infraction){
   switch infraction
-    case duplicate_vote: return duplicate_vote_rate
-    case light_client_attack: return light_client_attack
+    case "duplicate_vote": return duplicate_vote_rate
+    case "light_client_attack": return light_client_attack_rate
     default: panic()
 }
 ```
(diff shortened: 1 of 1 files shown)


2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -240,10 +248,12 @@ func bond(validator_address, delegator_address, amount)
...
     balances[pos] += amount
-    update_total_deltas(validator_address, pipeline_lenght, amount)
-    update_voting_power(validator_address, pipeline_lenght)
-    update_total_voting_power(pipeline_lenght)
-    update_validator_sets(validator_address, pipeline_lenght)
+    update_total_deltas(validator_address, pipeline_length, amount)
+    update_voting_power(validator_address, pipeline_length)
+    update_total_voting_power(pipeline_length)
+    // Only update validator sets if the validator is not jailed, since a jailed validator is not in a validator set
+    if !is_jailed(validator_address) then
```
angbrav: This is not needed, we remove a validator from active and inactive sets when jailing it.

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -292,10 +302,14 @@ func unbond(validator_address, delegator_address, total_amount)
...
         remain -= amount_unbonded
-      update_total_deltas(validator_address, pipeline_length, -1*amount_after_slashing)
+      
+      var pipeline_stake = read_epoched_field(validators[validator_address].total_deltas, cur_epoch + pipeline_length, 0)
+      var token_change = min{ amount_after_slashing, pipeline_stake }
+      update_total_deltas(validator_address, pipeline_length, -1*token_change)
       update_voting_power(validator_address, pipeline_length)
       update_total_voting_power(pipeline_length)
-      update_validator_sets(validator_address, pipeline_length)
+      if !is_jailed(validator_address) then
```
angbrav: Same as above: not needed.

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -320,15 +334,16 @@ func withdraw(validator_address, delegator_address)
...
-      //Update amount with slashes that happened more than `unbonding_length` before this slash
-      forall (slashed_amount in computed_amounts s.t. slashed_amount.epoch + unbonding_length < slash.epoch) do
-        updated_amount -= slashed_amount.amount
-        computed_amounts = computed_amounts \ {slashed_amount}
-      computed_amounts = computed_amounts \union {SlashedAmount{epoch: slash.epoch, amount: updated_amount*slash.rate}}
+  var updated_amount = amount
+  forall (slash in set_slashes in slash.epoch order) do
+    //Update amount with slashes that happened more than `unbonding_length` before this slash
+    forall (slashed_amount in computed_amounts s.t. slashed_amount.epoch + unbonding_length < slash.epoch) do
+      updated_amount = updated_amount.checked_sub(slashed_amount.amount)
```
angbrav: I would rather keep language specifics out. `checked_sub` is undefined within the spec. I guess you want to avoid `updated_amount` going negative. Maybe the more intuitive way of doing this is to simply break the loop if `update_amount` becomes negative and return 0 in that case.

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -292,10 +302,14 @@ func unbond(validator_address, delegator_address, total_amount)
         //(remove it, create a new one with double the amount, and add it).
         validators[validator_address].set_unbonds[cur_epoch+pipeline_length] = {UnbondRecord{amount: amount_unbonded, start: start}} \union validators[validator_address].set_unbonds[cur_epoch+pipeline_length]
         remain -= amount_unbonded
-      update_total_deltas(validator_address, pipeline_length, -1*amount_after_slashing)
+      
+      var pipeline_stake = read_epoched_field(validators[validator_address].total_deltas, cur_epoch + pipeline_length, 0)
+      var token_change = min{ amount_after_slashing, pipeline_stake }
+      update_total_deltas(validator_address, pipeline_length, -1*token_change)
```
angbrav: It looks to me that it would be enough to modify `compute_amount_after_slashing` to always return 0 or greater. If it returns 0, it means that the unbond is fully slashed. I believe the above lines are then not needed.

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -510,19 +524,21 @@ end_of_epoch()
 
   var set_validators = {val | val = slash.validator && slash in enqueued_slashes[cur_epoch]}
   forall (validator_address in set_validators) do
+    var total_staked = read_epoched_field(validators[validator_address].total_deltas, slash.epoch, 0)
```
angbrav: `slash` is undefined here, i.e, `slash.epoch` is as well. What's the motivation to move it out of the forall loop of slashes?

brentstone: I've changed `end_of_epoch()` again now in 61c733f to correct what I had done before and I think to solve the issue of `total_unbonded > total_staked`

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
brentstone: Addressed now in #50

angbrav (changes requested):
Thanks @brentstone! Everything looks good to me but the part that computes the final slash rate. We need to change `compute_final_rate` to implement what's described in https://specs.namada.net/economics/proof-of-stake/cubic-slashing.html. To mention one discrepancy, The pseudocode computes `slash.voting_power_fraction` (which is undefined and I assume that it is `stake_fraction`) by multiplying the misbehaving validator's stake at the misbehaving epoch to the infraction type rate. The spec does something different by dividing the validator's voting power by the total voting power (total PoS voting power)

Once we fix this, which should be easy, this PR can be merged.

brentstone:
> Thanks @brentstone! Everything looks good to me but the part that computes the final slash rate. We need to change `compute_final_rate` to implement what's described in https://specs.namada.net/economics/proof-of-stake/cubic-slashing.html. To mention one discrepancy, The pseudocode computes `slash.voting_power_fraction` (which is undefined and I assume that it is `stake_fraction`) by multiplying the misbehaving validator's stake at the misbehaving epoch to the infraction type rate. The spec does something different by dividing the validator's voting power by the total voting power (total PoS voting power)
> 
> Once we fix this, which should be easy, this PR can be merged.

Thanks for the note, I've addressed this in 48c13ba6e4475528111c100e980c50a9647f2c92. I've done some refactoring in order to sync this with the Namada spec and codebase. In particular, the `Slash` object has a field `voting_power` rather than `stake_fraction`, which indicates the voting power used to commit the infraction. As such, the function `compute_stake_fraction` is no longer needed and removed. I've also renamed `compute_final_rate` to `compute_cubic_rate`, as the final rate is computed later for each within the loop.

Lmk your thoughts.

angbrav:
I am still confused by the pseudocode in Namda's spec:
- Namada's spec computes the final rate as `min{1, max{get_min_slash_rate(slash), 9 * voting_power_fraction^2}}` (adapted to pseudocode model notation) and the pseudocode model does `max{get_min_slash_rate(slash), min{1, 9 * voting_power_fraction^2}}`, which is not equivalent. But I think there is something wrong in Namada's spec as `slash` is undefined. Is then what's currently in the pseudocode model the correct spec?
- It is unclear at which epoch the misbehaving validator's stake is retrieved in `validator.voting_power` by the Namada's spec. This is clearly underspecified in that pseudocode. The pseudocode model currently divides the validator's voting power at the misbehaving epoch (retrieved when the evidence is processed) by the total voting power at the current epoch. I am not sure that's intended. If it is, what's the intuition behind this?

Some minor thing. I would do something about
```go
func get_min_slash_rate(infraction){
  switch infraction
    case duplicate_vote: return duplicate_vote_rate
    case light_client_attack: return light_client_attack
    default: panic()
}
```
and add a field to the slash record with type of infraction for instance such that we call `get_min_slash_rate` for a `slash` by `get_min_slash_rate(slash.type)` instead of `get_min_slash_rate(slash)`.

brentstone:
> I am still confused by the pseudocode in Namda's spec:
> 
> * Namada's spec computes the final rate as `min{1, max{get_min_slash_rate(slash), 9 * voting_power_fraction^2}}` (adapted to pseudocode model notation) and the pseudocode model does `max{get_min_slash_rate(slash), min{1, 9 * voting_power_fraction^2}}`, which is not equivalent. But I think there is something wrong in Namada's spec as `slash` is undefined. Is then what's currently in the pseudocode model the correct spec?
> * It is unclear at which epoch the misbehaving validator's stake is retrieved in `validator.voting_power` by the Namada's spec. This is clearly underspecified in that pseudocode. The pseudocode model currently divides the validator's voting power at the misbehaving epoch (retrieved when the evidence is processed) by the total voting power at the current epoch. I am not sure that's intended. If it is, what's the intuition behind this?
> 
> Some minor thing. I would do something about
> 
> ```go
> func get_min_slash_rate(infraction){
>   switch infraction
>     case duplicate_vote: return duplicate_vote_rate
>     case light_client_attack: return light_client_attack
>     default: panic()
> }
> ```
> 
> and add a field to the slash record with type of infraction for instance such that we call `get_min_slash_rate` for a `slash` by `get_min_slash_rate(slash.type)` instead of `get_min_slash_rate(slash)`.

Yeah, so first, what is in the pseudocode is more precise in terms of specifying the actual code of Namada. The Namada spec for cubic slashing was also updated today a bit, but nevertheless is a bit more general than an approach that walks through the code very precisely. Anyway, will make some changes to the Namada spec as there are currently some rendering issues and things can be specified more precisely.

You are right that I have a bug in the epochs used. Thanks, I've fixed this. Also have integrated your suggested change regarding `get_min_slash_rate`. See the changes in 6dfa21e.

brentstone:
Since you brought up certain things that were undefined as well, I also notice that `evidence` with the fields `validator` and `epoch` may be undefined as well.

angbrav:
> Since you brought up certain things that were undefined as well, I also notice that `evidence` with the fields `validator` and `epoch` may be undefined as well.

I would simply assume that `new_evidence` has `type`, `epoch` and `validator` as arguments, instead of the undefined `evidence` struct`.

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -345,7 +347,7 @@ func new_evidence(evidence)
 {
   //create slash
   var total_staked = read_epoched_field(validators[evidence.validator].total_deltas, evidence.epoch, 0)
-  var slash = Slash{epoch: evidence.epoch, validator: evidence.validator, rate: 0, stake_fraction: compute_stake_fraction(evidence.type, total_staked)}
+  var slash = Slash{epoch: evidence.epoch, validator: evidence.validator, rate: 0, voting_power: total_staked}
```
angbrav: See [my comment](https://github.com/informalsystems/partnership-heliax/pull/43#issuecomment-1560660573), it is probably better to remove `evidence` struct, hence my suggestion.
```suggestion
  var slash = Slash{epoch: epoch, validator: validator, rate: 0, type: type, voting_power: total_staked}
```

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -54,7 +54,8 @@ type Slash struct {
   epoch Epoch
   validator Addr
   rate float
-  stake_fraction float //new in cubic slashing
+  voting_power VotingPower // the voting power used to commit the infraction
+  slash_type string
```
angbrav:
```suggestion
  type string
```

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -552,13 +554,25 @@ end_of_epoch()
...
 
 ```go
-//Cubic slashing function
-compute_final_rate(slashes)
+// Get total stake of active (consensus-participating) validators
+get_total_active_voting_power(epoch)
 {
+  var voting_power = 0
+  forall (validator in validator_sets[epoch].active)
+    voting_power += validator.voting_power
```
angbrav:
```suggestion
    voting_power += `read_epoched_field(validators[validator].voting_power, epoch, 0)`
```
I know we are using `validator.voting_power` in some other places, e.g., `update_total_voting_power` or `update_validator_sets`, but I think we should not. We can fix the other functions in a different PR.

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -552,13 +554,25 @@ end_of_epoch()
 ```
 
 ```go
-//Cubic slashing function
-compute_final_rate(slashes)
+// Get total stake of active (consensus-participating) validators
+get_total_active_voting_power(epoch)
```
angbrav, line 558: It may be useful to precompute both the total active voting power per epoch and the total inactive voting power, as opposed to what we do now: to precomputing the sum of both in `update_total_voting_power`.

angbrav (commented):
I have left a bunch of comments, but nothing serious.

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
brentstone: perhaps this can be in another PR where we make updates as suggested in https://github.com/informalsystems/partnership-heliax/pull/43#discussion_r1203744479?

angbrav (approved):
LGTM, let's merge it, thanks!

PR #45, opened 2023-04-11
Add redelegation to the pseudocode model

angbrav:
This PR adds redelegation to the pseudocode model and English text to describe the main logic and some definitions.

Note that the text below may be a bit outdatet.

### Definitions
- We use source and destination validators to refer to the validator from which the tokens are redelegation and the validator to which the tokens are redelegated respectively.
- We say that a redelegation starts the epoch at which the `tx_redelegate` tx is issued and ends when the redelegated tokens stop contributing to the source validator and start contributing to the destination validator.
- Chain redelegations occur when some redelegated tokens are redelegated within less than `unbonding_length` epochs.

### Desired Properties
The solution aim to guarantee the following desired properties:
- **Fast redelegation**: Redelegated tokens start contributing to the destination validator stake after `pipeline_length` epochs. This means that if a delegator redelegates X tokens at epoch `e`, then these tokens stop contributing to the source validator stake and start contributing to the destination validator stake at `e+pipeline_length`.
- **Minimize changes**: Adding redelegation should not require a major redesign of the existing protocol.

### Data Structures
The solution introduces the following data structures to track redelegations:
- `Redelegation` is a record with two fields: the source validator and the amount of tokens redelegated.
- `redelegated_bonds in (Addr X Addr X Epoch) → Redelegation`: A map from delegator to a map from redelegation's destination validator to a map from the epoch at which the redelegation ends to a Redelegation record. This sounds more complex than it is. This is accessed by simply `redelegated_bonds[delegator][dest_validator][epoch]`. It is the counterparty of the `bonds` data structure. 
- `redelegated_unbonds[][][] in (Addr X Addr X (Epoch, Epoch)) → Redelegation`: Similar to `redelegated_bonds` but the key of the last map is a tuple to track not only the end of the redelegation but also the when the tokens are unbonded at the destination validator. It is the counterparty of the `unbonds` data structure.
- Each validator now keeps track of incoming validators in a map: `redelegations map<Addr, map<Addr, IncomingRedelegation>>`. This is useful to prevent redelegations. `IncomingRedelegation` is a record that stores the amount redelegated and the epoch at which the redelegation started.

### How it roughly works
- The main complexity comes from the fact that we do not keep a separate record for each bond/unbond but we accumulate them per epoch (epochs for unbonds) in a delta data structure. This makes things complicated because when slashing we need to know the precise amount of tokens that come from a redelegation to be able to apply slashes of the source validator if required. `redelegated_bonds` and `redelegated_unbonds` aim at solving this issue. Let's take the following example:
  - Assume a delegator `del`  and two validators `val1` and `val2`.
  - `del` redelegates X tokens from `val1` to `val2` in epoch `e`. Then the redelegation ends at `e + pipeline_length`. Then `bonds[del][val2].deltas[e+pipeline_length]=X`, assuming that there were no bonds before.
  - `del` delegates `Y` tokens to `val2` in epoch `e` as well. Then `bonds[del][val2].deltas[e+pipeline_length]=X+Y`.
  - Assume now that `del` want to unbond at an epoch `e' >= e+pipeline_length` and the bonds in `bonds[del][val2].deltas[e+pipeline_length]` are going to be unbonded. We need to do two things:
    - Apply all slashes processed by `e'` that `val2` committed in any epoch ` >= e+ pipeline_length` to the whole amount (X+Y).
    - Also, we need to apply any slashes that the source validator committed while the Y tokens were being redelegated (any slash not processed before the redelegation started and committed before the redelegation ends.
  - To compute X and Y precisely, the solution relies on `redelegated_bonds`. Thus:
    - `bonds[del][val2].deltas[e+pipeline_length]=X+Y`
    - `redelegated_bonds[del][val2][e+pipeline_length]=Y`
    - X can be computed from the above.
- The above logic has to be applied not only on unbonding, but also when withdrawing and at the end of an epoch when a validator is slashed.
- Other details:
  - When we redelegate tokens, we do not just simply move the bond amounts to the destination validator, but we slash first. Thus, the bonded amount at the destination validator already accounts for all the slashes committed by the source validator processed before the redelegation starts. This means that we only need to care about slashes of the source validator that were processed after the redelegation started and before it ended (at that point, the tokens do not contribute to the source validator stake).
  - Another tricky part is what happens at the end of an epoch. We first need to slash the misbehaving validator. Then we have to slash any validator for which the misbehaving validator redelegated tokens to. Slashes are always filtered based on when the infraction was committed and when the redelegation started/ended.

### DIFF
--- a/2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
+++ b/2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
@@ -280,8 +280,7 @@ The transaction `tx_redelegate` creates a redelegation from source validator `sr
 1. The function first checks that the source validator is not frozen. This is to prevent unbonding at an epoch `e` from a validator that it is known to have behaved, but the slash has not been processed yet.
 2. Then it checks that the redelegation is not a chain redelegation, in which case it prevents it by returning.
    - The design does not track chain redelegation precisely.
-   - Instead approximates it by preventing a validator to redelegate tokens belonging to a delegator if it has recently received a redelegation of tokens belonging to the same delegator.
-   - The `incoming_redelegations` variable is used to implement this mechanism.
+   - Instead approximates it by preventing a validator to redelegate tokens belonging to a delegator if it has received a redelegation of tokens belonging to the same delegator that ended `unbonding_length` epochs ago. Please read the definition of chain redelegation in [Definitions](#definitions) for more details.
 3. The function computes the total amount of tokens that the delegator has bonded at the source validator (`bonded_tokens`).
 4. It unbonds `bonded_tokens` from the source validator indicating that it is a redelegation by calling `unbond` with the delegator's address.
 5. It creates the bond at the destination validator.
--- a/2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
+++ b/2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
@@ -634,7 +634,7 @@ end_of_epoch()
       slash.rate = rate
       append(slashes[validator_address], slash)
       var total_staked = read_epoched_field(validators[validator_address].total_deltas, slash.epoch, 0)
-      slash_misbehaving_validator(validator_address, slash, staked_amount)
+      slash_misbehaving_validator(validator_address, slash, total_staked)
 
       forall (val in validators) do
         var pairs = { pair = <delegator, <epoch, redelegation> | pair in validators[val].redelegations[validator_address] &&
--- a/2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
+++ b/2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
@@ -219,10 +219,10 @@ tx_undelegate(validator_address, delegator_address, amount)
 ```
 
 ```go
-func is_chained_redelegation(dest_validator_address, delegator_address)
+func is_chained_redelegation(validator_address, delegator_address)
   // avoid iteration over validators and do it over incoming_redelegations
   forall (val in validators) do
-    var redelegation = validators[dest_validator_address].redelegations[val][delegator_address]
+    var redelegation = validators[validator_address].redelegations[val][delegator_address]
     if (redelegation.epoch >= 0 && redelegation.epoch + unbonding_length > cur_epoch) then
       return true
   return false
@@ -234,7 +234,7 @@ tx_redelegate(src_validator_address, dest_validator_address, delegator_address)
   // Disallow re-delegation if the source validator is frozen (similar to unbonding)
   var src_frozen = read_epoched_field(validators[src_validator_address].frozen, cur_epoch, false)
   if (is_validator(src_validator_address, cur_epoch+pipeline_length) && src_frozen == false) then
-    // Check that `incoming_redelegations[delegator_address]` for `dest_validator_address` either don't exist
+    // Check that `incoming_redelegations[delegator_address]` for `src_validator_address` either don't exist
     // or if they do, they cannot be slashed anymore (`end + unbonding_length <= cur_epoch`)
     if is_chained_redelegation(src_validator_address, delegator_address) then
       return
--- a/2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
+++ b/2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
@@ -57,6 +57,14 @@ type Slash struct {
   stake_fraction float //new in cubic slashing
 }
 
+type Redelegation struct {
+  start Epoch
+  end Epoch
+  unbonded Epoch
+  validator Addr
+  amount uint
+}
+
 type WeightedValidator struct {
   validator Addr
   voting_power VotingPower
@@ -204,15 +212,44 @@ tx_undelegate(validator_address, delegator_address, amount)
 }
 ```
 
-<!-- 
 ```go
-tx_redelegate(src_validator_address, dest_validator_address, delegator_address, amount)
+func is_chained_redelegation(dest_validator_address, delegator_address)
+  // avoid iteration over validators and do it over incoming_redelegations
+  forall (val in validators) do
+    var incoming = validators[dest_validator_address].incoming_redelegations[val][delegator_address]
+    if (incoming != ⊥ && incoming.end + unbonding_length > cur_epoch) then
+      return true
+  return false
+```
+
+```go
+tx_redelegate(src_validator_address, dest_validator_address, delegator_address)
 {
-  unbond(src_validator_address, delegator_address, amount)
-  bond(dst_validator_address, delegator_address, amount, unbonding_length)
+  // Disallow re-delegation if the source validator is frozen (similar to unbonding)
+  var src_frozen = read_epoched_field(validators[src_validator_address].frozen, cur_epoch, false)
+  if (is_validator(src_validator_address, cur_epoch+pipeline_length) && src_frozen == false) then
+    // Check that `incoming_redelegations[delegator_address]` for `dest_validator_address` either don't exist
+    // or if they do, they cannot be slashed anymore (`end + unbonding_length <= cur_epoch`)
+    if is_chained_redelegation(dest_validator_address, delegator_address) then
+      return
+    // Find the sum of bonded tokens to `src_validator_address`
+    var delbonds = {<start, amount> | amount = bonds[delegator_address][src_validator_address].deltas[start] > 0 && start <= cur_epoch + unbonding_length}
+    var bonded_tokens = sum{amount | <start, amount> in delbonds}
+    // Unbond the tokens from `src_validator_address` at pipeline offset, but 
+    // without recording it in the `unbonds[delegator_address][src_validator_address]`
+    var record_unbonds = false
+    var amount_after_slashing = unbond(src_validator_address, delegator_address, bonded_tokens, record_unbonds)
+    // add a bond in the `dest_validator_address` for the amount after slashing
+    // done manually to avoid account transfer
+    bonds[delegator_address][dest_validator_address].deltas[cur_epoch+pipeline_length] += amount_after_slashing
+    bonds_redelegations[delegator_address][dest_validator_address][cur_epoch+pipeline_length] = RedelegatedBond{validator: src_validator_address,
+                                                                                                                amount: amount_after_slashing}
+    update_total_deltas(dest_validator_address, pipeline_length, amount_after_slashing)
+    update_voting_power(dest_validator_address, pipeline_length)
+    update_total_voting_power(pipeline_length)
+    update_validator_sets(dest_validator_address, pipeline_length)
 }
 ```
--->
 
 ```go
 tx_withdraw_unbonds_delegator(delegator_address)
@@ -223,7 +260,6 @@ tx_withdraw_unbonds_delegator(delegator_address)
 ```
 
 ## PoS functions
-
 
 ```go
 //This function is called by transactions tx_self_bond, tx_delegate and tx_redelegate
@@ -261,7 +297,7 @@ func bond(validator_address, delegator_address, amount)
     conclusion: it is an actual issue, unresolved
 */
 //This function is called by transactions tx_unbond, tx_undelegate and tx_redelegate
-func unbond(validator_address, delegator_address, total_amount)
+func unbond(validator_address, delegator_address, total_amount, record_unbonds)
 {
   //disallow unbonding if the validator is frozen
   var frozen = read_epoched_field(validators[validator_address].frozen, cur_epoch, false)
@@ -272,16 +308,49 @@ func unbond(validator_address, delegator_address, total_amount)
     //this serves to check that there are bonds (in the docs) and that these are greater than the amount we are trying to unbond
     if (sum{amount | <start, amount> in delbonds} >= total_amount) then
       var remain = total_amount
-      var amount_after_slashing = total_amount
+      var amount_after_slashing = 0
       //Iterate over bonds and create unbond
       forall (<start, amount> in delbonds while remain > 0) do
+        //Retrieve the bond's associated redelegation record (if any)
+        var redelegation_record = redelegations[delegator_address][dest_validator_address][start]
         //Take the minimum between the remainder and the unbond. This is equal to amount if remain > amount and remain otherwise 
         var amount_unbonded = min{amount, remain}
+        //if there is redelegation record and the redelegated amount is less than the bond amount,
+        //then compute slashes separately: for redelegated and non-redelegated amount
+        if redelegation_record != ⊥ && redelegation_record.amount < amount_unbonded then
+          // set of slashes that happened while the bond was contributing to the validator's stake
+          var set_slashes = {s | s in slashes[validator_address] && start <= slash.epoch }
+          var amount_unbonded_after_slashing = compute_amount_after_slashing(set_slashes, amount_unbonded - redelegation_record.amount)
+          // add slashes if the tokens were redelegated
+          set_slashes = set_slashes \union slashes_due_redelegation(redelegation_record, start)
+          amount_unbonded_after_slashing += compute_amount_after_slashing(set_slashes, redelegation_record.amount)
+        else
+          // set of slashes that happened while the bond was contributing to the validator's stake
+          var set_slashes = {s | s in slashes[validator_address] && start <= slash.epoch }
+          // add slashes if the tokens were redelegated
+          set_slashes = set_slashes \union slashes_due_redelegation(redelegation_record)
+          var amount_unbonded_after_slashing = compute_amount_after_slashing(set_slashes, amount_unbonded)
+      
+        //update bond and create new unbond if record_unbonds=true
         bonds[delegator_address][validator_address].deltas[start] = amount - amount_unbonded
-        unbonds[delegator_address][validator_address].deltas[start, cur_epoch+pipeline_length+unbonding_length] = amount_unbonded
-        // set of slashes that happened while the bond was contributing to the validator's stake
-        var set_slashes = {s | s in slashes[validator_address] && start <= slash.epoch }
-        amount_after_slashing += compute_amount_after_slashing(set_slashes, amount_unbonded)
+
+        var end = cur_epoch+pipeline_length+unbonding_length
+        if record_unbonds then
+          unbonds[delegator_address][validator_address].deltas[start, end] = amount_unbonded
+
+        //manage redelegation records
+        if redelegation_record != ⊥ then
+          if record_unbonds then
+            //TODO: there may exists a record already
+            unbonds_redelegations[delegator_address][dest_validator_address][start, end] = Redelegation{validator: src_validator_address,
+                                                                                                        redelegation_record.amount - amount_unbonded}
+          if redelegation_record.amount > amount_unbonded then
+            bonds_redelegations[delegator_address][dest_validator_address][start] = Redelegation{validator: src_validator_address,
+                                                                                                 redelegation_record.amount - amount_unbonded}
+          else
+            bonds_redelegations[delegator_address][dest_validator_address][start] = ⊥
+  
+        amount_after_slashing += amount_unbonded_after_slashing
         //The current model disregards a corner case that should be taken care of in the implementation:
         //- Assume a user delegates 10 tokens to a validator at epoch e1
         //- Assume the user unbonds twice 5 tokens from the same validator in the same epoch in two different transactions e2
@@ -296,6 +365,20 @@ func unbond(validator_address, delegator_address, total_amount)
       update_voting_power(validator_address, pipeline_length)
       update_total_voting_power(pipeline_length)
       update_validator_sets(validator_address, pipeline_length)
+      return amount_after_slashing
+}
+```
+
+```go
+func slashes_due_redelegation(redelegation_record, end_redelegation)
+{
+  if redelegation_record == ⊥ then
+    return {}
+  //all slashes for the source validator while the tokens were delegated it
+  return {slash | slash in slashes[redelegation_record.validator] &&
+                  //redelegation_record.end - pipeline_length is the epoch when the redelegation was issued
+                  //any slashed processed at an epoch < redelegation_record.end - pipeline_length has been applied already
+                  end_redelegation - pipeline_length - unbonding_length <= slash.epoch < end_redelegation}
 }
 ```
 
@@ -306,9 +389,10 @@ func withdraw(validator_address, delegator_address)
 {
   var delunbonds = {<start,end,amount> | amount = unbonds[delegator_address][validator_address].deltas[(start, end)] > 0 && end <= cur_epoch }
   //substract any pending slash before withdrawing
-  forall (<start,end,amount> in selfunbonds) do
+  forall (<start,end,amount,redelegation_record> in selfunbonds) do
     // set of slashes that happened while the bond associated with the unbond was contributing to the validator's stake
     var set_slashes = {s | s in slashes[validator_address] && start <= slash.epoch && slash.epoch < end - unbonding_length }
+    set_slashes = set_slashes \union slashes_due_redelegation(redelegation_record)
     var amount_after_slashing = compute_amount_after_slashing(set_slashes, amount)
     balance[delegator_address] += amount_after_slashing
     balance[pos] -= amount_after_slashing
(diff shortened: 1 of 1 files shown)


2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -204,15 +218,47 @@ tx_undelegate(validator_address, delegator_address, amount)
...
+    var record_unbonds = false
+    var amount_after_slashing = unbond(src_validator_address, delegator_address, bonded_tokens, record_unbonds)
+    // add a bond in the `dest_validator_address` for the amount after slashing
+    // done manually to avoid account transfer
+    bonds[delegator_address][dest_validator_address].deltas[cur_epoch+pipeline_length] += amount_after_slashing
+    redelegated_bonds[delegator_address][dest_validator_address][cur_epoch+pipeline_length] = Redelegation{validator: src_validator_address,
+                                                                                                           amount: amount_after_slashing}
+    //save the epoch at which the redelegation occurs to track chained redelegations
+    validators[dest_validator_address].redelegations[src_validator_address][delegator_address] = IncomingRedelegation{epoch: cur_epoch,
+                                                                                                                      amoun: amount_after_slashing}
```
tzemanovic:
```suggestion
                                                                                                                      amount: amount_after_slashing}
```

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -290,35 +377,67 @@ func unbond(validator_address, delegator_address, total_amount)
...
 //This function is called by transactions tx_withdraw_unbonds_validator and tx_withdraw_unbonds_delegator
 func withdraw(validator_address, delegator_address)
 {
   var delunbonds = {<start,end,amount> | amount = unbonds[delegator_address][validator_address].deltas[(start, end)] > 0 && end <= cur_epoch }
   //substract any pending slash before withdrawing
-  forall (<start,end,amount> in selfunbonds) do
-    // set of slashes that happened while the bond associated with the unbond was contributing to the validator's stake
-    var set_slashes = {s | s in slashes[validator_address] && start <= slash.epoch && slash.epoch < end - unbonding_length }
-    var amount_after_slashing = compute_amount_after_slashing(set_slashes, amount)
+  forall (<start,end,amount,redelegation_record> in selfunbonds) do
```
tzemanovic:
```suggestion
  forall (<start,end,amount,redelegation_record> in delunbonds) do
```

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -503,44 +622,28 @@ func compute_stake_fraction(infraction, voting_power){
...
-        var this_slash = (total_staked - total_unbonded) * slash.rate
-        var diff_slashed_amount = last_slash - this_slash
-        last_slash = this_slash
-        update_total_deltas(validator_address, offset, diff_slashed_amount)
-        update_voting_power(validator_address, offset)
-        total_unbonded = 0
+      slash_misbehaving_validator(validator_address, slash, staked_amount)
+
+      forall (val in validators) do
+        var pairs = { pair = <delegator, <epoch, redelegation> | pair in validators[var].redelegations[validator_address] &&
```
tzemanovic:
```suggestion
        var pairs = { pair = <delegator, <epoch, redelegation> | pair in validators[val].redelegations[validator_address] &&
```

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -550,6 +653,112 @@ end_of_epoch()
...
+      // We only need to apply a slash s if s.epoch < unbond.end - unbonding_length
+      // It is easy to see that s.epoch + unbonding_length < slash.epoch implies s.epoch < unbond.end - unbonding_length
+      // 1) slash.epoch = cur_epoch - unbonding_length
+      // 2) unbond.end = cur_epoch + offset + unbonding_length => cur_epoch = unbond.end - offset - unbonding_length
+      // By 1) s.epoch + unbonding_length < cur_epoch - unbonding_length
+      // By 2) s.epoch + unbonding_length < unbond.end - offset - 2*unbonding_length => s.epoch < unbond.end - offset - 3*unbonding_length, as required.
+      var set_slashes = {s | s in slashes[validator_address] &&
+                             unbond.start <= s.epoch &&
+                             s.epoch + unbonding_length < slash.epoch }
+      var total_unbonded = compute_amount_after_slashing(set_slashes, unbond.amount - redelegated_amount)
```
tzemanovic:
```suggestion
     total_unbonded += compute_amount_after_slashing(set_slashes, unbond.amount - redelegated_amount)
```

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -550,6 +653,112 @@ end_of_epoch()
 }
 ```
 
+```go
+func slash_misbehaving_validator(src_validator_address, dest_validator_address, slash, amount)
```
tzemanovic:
```suggestion
func slash_misbehaving_validator(src_validator_address, dest_validator_address, slash, total_staked)
```

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -503,44 +622,28 @@ func compute_stake_fraction(infraction, voting_power){
...
-          // By 2) s.epoch + unbonding_length < unbond.end - offset - 2*unbonding_length => s.epoch < unbond.end - offset - 3*unbonding_length, as required.
-          var set_prev_slashes = {s | s in slashes[validator_address] && unbond.start <= s.epoch && s.epoch + unbonding_length < slash.epoch}
-          total_unbonded += compute_amount_after_slashing(set_prev_slashes, unbond.amount)
-        var this_slash = (total_staked - total_unbonded) * slash.rate
-        var diff_slashed_amount = last_slash - this_slash
-        last_slash = this_slash
-        update_total_deltas(validator_address, offset, diff_slashed_amount)
-        update_voting_power(validator_address, offset)
-        total_unbonded = 0
+      slash_misbehaving_validator(validator_address, slash, staked_amount)
```
tzemanovic:
```suggestion
      slash_misbehaving_validator(validator_address, slash, total_staked)
```

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -550,6 +653,112 @@ end_of_epoch()
...
+                                              //slash s is processed before the misbehaving epoch (slash.epoch)
+                                              s.epoch + unbonding_length < slash.epoch}
+        total_unbonded += compute_amount_after_slashing(set_slashes, redelegated_amount)
+
+  var last_slash = 0
+  // up to pipeline_length because there cannot be any unbond in a greater ß (cur_epoch+pipeline_length is the upper bound)
+  forall (offset in 1..pipeline_length) do
+    forall (unbond in validators[validator_address].set_unbonds[cur_epoch + offset] s.t. unbond.start <= slash.epoch) do
+
+    //retrieve redelegation record
```
tzemanovic: I think the lines from here down to line 714 (`total_unbonded += ...`) are misaligned and should be inside the inner for loop

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -204,15 +218,47 @@ tx_undelegate(validator_address, delegator_address, amount)
...
+tx_redelegate(src_validator_address, dest_validator_address, delegator_address)
 {
-  unbond(src_validator_address, delegator_address, amount)
-  bond(dst_validator_address, delegator_address, amount, unbonding_length)
+  // Disallow re-delegation if the source validator is frozen (similar to unbonding)
+  var src_frozen = read_epoched_field(validators[src_validator_address].frozen, cur_epoch, false)
+  if (is_validator(src_validator_address, cur_epoch+pipeline_length) && src_frozen == false) then
+    // Check that `incoming_redelegations[delegator_address]` for `dest_validator_address` either don't exist
+    // or if they do, they cannot be slashed anymore (`end + unbonding_length <= cur_epoch`)
+    if is_chained_redelegation(dest_validator_address, delegator_address) then
```
tzemanovic: do we need to also check it for the src validator? i.e. `is_chained_redelegation(src_validator_address, delegator_address)`

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -204,15 +218,47 @@ tx_undelegate(validator_address, delegator_address, amount)
...
+  // avoid iteration over validators and do it over incoming_redelegations
+  forall (val in validators) do
+    var redelegation = validators[dest_validator_address].redelegations[val][delegator_address]
+    if (redelegation.epoch >= 0 && redelegation.epoch + unbonding_length > cur_epoch) then
+      return true
+  return false
+```
+
+```go
+tx_redelegate(src_validator_address, dest_validator_address, delegator_address)
```
tzemanovic, line 294:
```suggestion
func tx_redelegate(src_validator_address, dest_validator_address, delegator_address)
```

cwgoes:
(ref some discussions in https://github.com/anoma/namada/issues/34)

We need to make sure not to iterate over delegations or redelegations (DoS vector). In this case I think we can treat all redelegations from X to Y in epoch e the same (so e.g. when slashing we can just store the total amount from X to Y in e and slash based on this, we don't have to iterate over the redelegations).

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
tzemanovic: as discussed this should rather be
```suggestion
    if is_chained_redelegation(src_validator_address, delegator_address) then
```

tzemanovic:
> (ref some discussions in [anoma/namada#34](https://github.com/anoma/namada/issues/34))
> 
> We need to make sure not to iterate over delegations or redelegations (DoS vector). In this case I think we can treat all redelegations from X to Y in epoch e the same (so e.g. when slashing we can just store the total amount from X to Y in e and slash based on this, we don't have to iterate over the redelegations).

Thanks, that's a very important point! I think we unwittingly introduced iteration in https://github.com/informalsystems/partnership-heliax/pull/38 that's being carried over to here too, where in `end_of_epoch` we started iterating over `validator.set_unbonds`.

For the redelegations themselves, we suggested grouping these by a triple of src validator, dest validator and redelegation epoch, and then iteration over the groups is acceptable.

I think we could solve the iteration over the `set_unbonds` the same way - i.e. group unbonds by a pair of bond start epoch and unbond epoch.

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -272,16 +317,58 @@ func unbond(validator_address, delegator_address, total_amount)
...
+        if redelegated_amount < amount_unbonded then 
+          // set of slashes that happened while the bond was contributing to the validator's stake
+          var set_slashes = {s | s in slashes[validator_address] && start <= s.epoch }
+          amount_unbonded_after_slashing = compute_amount_after_slashing(set_slashes, amount_unbonded - redelegated_amount)
+        //compute amount after slashing  of the tokens coming from redelegation
+        if redelegation_record != ⊥ then
+          set_slashes = set_slashes \union {slash | slash in slashes[redelegation_record.validator] &&
+                                                    //start - pipeline_length is the epoch when the redelegation was issued
+                                                    //any slashed processed at an epoch < slash - pipeline_length - unbonding_length has been applied already
+                                                    start - pipeline_length - unbonding_length <= slash.epoch < start}
```
tzemanovic: If I understand correctly, for a re-delegation we cannot determine the start epoch on the source validator, so we're just applying the slash indiscriminately as long as it's in the slashable duration here, right?

I don't think it's feasible to apply these exactly unless we do something even more complicated, as the re-delegation might be created from multiple bond deltas each with different start epoch, so I think it's is a good approximation, but would like to confirm and document this.

tzemanovic: This ofc applies to the replicated logic in withdraw and end_of_epoch slashing

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -550,6 +653,112 @@ end_of_epoch()
...
+    //filter those unbonds that were contributing to the stake when the validator misbehaved
+    forall (unbond in validators[validator_address].set_unbonds[epoch] s.t. unbond.start - unbonding_length - pipeline_length <= slash.epoch && 
+                                                                            unbond.redelegation != ⊥ &&
+                                                                            unbond.redelegation.validator == src_validator) do
+      //TODO: check that this set is not always empty
+      var set_slashes = {s | s in slashes[validator_address] &&
+                             unbond.start - unbonding_length - pipeline_length <= s.epoch &&
+                             s.epoch + unbonding_length < slash.epoch }
+
+      total_unbonded = compute_amount_after_slashing(set_slashes, unbond.amount)
```
tzemanovic:
```suggestion
      total_unbonded += compute_amount_after_slashing(set_slashes, unbond.amount)
```

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -272,16 +317,58 @@ func unbond(validator_address, delegator_address, total_amount)
...
+        var redelegation_record_set_unbonds = ⊥
+        if redelegation_record != ⊥ then
+          
+          if record_unbonds then
+            update_unbond_redelegation_record(delegator_address,
+                                              validator_address,
+                                              redelegation_record.validator,
+                                              start,
+                                              end,
+                                              min{redelegation_record.amount, amount_unbonded})
```
tzemanovic: I think we need to negate the amount here to subtract it from the record, right?

tzemanovic: ah, nvm, this is the redelegation unbond

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -503,44 +622,28 @@ func compute_stake_fraction(infraction, voting_power){
...
-          total_unbonded += compute_amount_after_slashing(set_prev_slashes, unbond.amount)
-        var this_slash = (total_staked - total_unbonded) * slash.rate
-        var diff_slashed_amount = last_slash - this_slash
-        last_slash = this_slash
-        update_total_deltas(validator_address, offset, diff_slashed_amount)
-        update_voting_power(validator_address, offset)
-        total_unbonded = 0
+      slash_misbehaving_validator(validator_address, slash, total_staked)
+
+      forall (val in validators) do
```
tzemanovic: I think we'll want to avoid looping over all validators here to find the outgoing redelegations for `validator_address`. We could additionally record outgoing redelegations in `tx_redelegate` to make this a look-up without iteration.

To consider the DoS vector (https://github.com/informalsystems/partnership-heliax/pull/45#issuecomment-1519447279), we don't need to know the `delegator` address for these and so we can also group these by a triple of src validator, dest validator and redelegation epoch, e.g.:

```
outgoing_redelegations[src_validator_address][dest_validator_address][cur_epoch+pipeline_length] += amount_after_slashing
```

and then in here we should be able to iterate over `outgoing_redelegations[validator_address]` within the applicable epoch window

tzemanovic:
I tried to check for these properties:

1. bonding, unbonding, withdrawal and slashing from a validator that has no outgoing or incoming re-delegations is unaffected by the changes
2. when a re-delegation src validator is slashed before re-delegation epoch + pipeline and the src bond was in slashable epoch range, the slash for the re-delegated amount is subtracted for the slash applied on the src validator and applied on the dest validator instead
3. when a re-delegation src validator is slashed at or after re-delegation epoch + pipeline, the dest validator is unaffected
4. when a re-delegation dest validator is slashed before re-delegation epoch + pipeline, the re-delegated amount is unaffected
5. when a re-delegation dest validator is slashed at or after re-delegation epoch + pipeline, the re-delegated amount is slashed
6. when a re-delegation is unbonded after a slash on the src validator that's applicable to the re-delegation is discovered, the unbond gets slashed and the final amount that can be withdrawn is not slashed again
7. when a re-delegation is unbonded before a slash on the src validator that's applicable to the re-delegation is discovered, the unbond isn't slashed and the final amount that can be withdrawn gets slashed
8. when a re-delegation is unbonded and a slash on the dest validator that's applicable to the re-delegation is discovered, the unbond isn't slashed and the final amount that can be withdrawn gets slashed (like a regular bond unbond)

I haven't found any issues with the logic besides the comments, most of which are minor. 

As per https://github.com/informalsystems/partnership-heliax/pull/45#discussion_r1174870619 I think for 2., 6. and 7. we're ignoring the start epochs of the src bond deltas, which I think is ok as we never skip a slash on something that should be slashed, only potentially slashing a re-delegation that might not need to be slashed if we were to track the start epochs of re-delegations in detail from the source deltas. I think this is applied consistently across the different paths (unbond/withdraw/end of epoch).

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
angbrav: I am not using `func` for txs for some reason (see the rest).

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
angbrav: > If I understand correctly, for a re-delegation we cannot determine the start epoch on the source validator, so we're just applying the slash indiscriminately as long as it's in the slashable duration here, right?

> I don't think it's feasible to apply these exactly unless we do something even more complicated, as the re-delegation might be created from multiple bond deltas each with different start epoch, so I think it's is a good approximation, but would like to confirm and document this.

We do not need to know when it started because we are slashing when doing the redelegation. Thus, we are not doing any approximation. Let me elaborate a bit:
- When the redelegation is issued, we compute the redelegated amount after slashing the source validator.
- If the redegelation happens at `e`, this means that we have processed all slashes up to `e-unbonding_lengh` (not included): an infraction at an epoch is processed at the end of that epoch + `unbonding_length`.
- Thus, when ubonding, we need to apply slashes for the source validator from `e-unbonding_lengh` to `e+pipeline_lentgh` (not included): when the redelegation ends.

Does this clarify it?

angbrav:
> I tried to check for these properties:
> 
> 1. bonding, unbonding, withdrawal and slashing from a validator that has no outgoing or incoming re-delegations is unaffected by the changes
> 2. when a re-delegation src validator is slashed before re-delegation epoch + pipeline and the src bond was in slashable epoch range, the slash for the re-delegated amount is subtracted for the slash applied on the src validator and applied on the dest validator instead
> 3. when a re-delegation src validator is slashed at or after re-delegation epoch + pipeline, the dest validator is unaffected
> 4. when a re-delegation dest validator is slashed before re-delegation epoch + pipeline, the re-delegated amount is unaffected
> 5. when a re-delegation dest validator is slashed at or after re-delegation epoch + pipeline, the re-delegated amount is slashed
> 6. when a re-delegation is unbonded after a slash on the src validator that's applicable to the re-delegation is discovered, the unbond gets slashed and the final amount that can be withdrawn is not slashed again
> 7. when a re-delegation is unbonded before a slash on the src validator that's applicable to the re-delegation is discovered, the unbond isn't slashed and the final amount that can be withdrawn gets slashed
> 8. when a re-delegation is unbonded and a slash on the dest validator that's applicable to the re-delegation is discovered, the unbond isn't slashed and the final amount that can be withdrawn gets slashed (like a regular bond unbond)
> 
> I haven't found any issues with the logic besides the comments, most of which are minor.
> 
> As per [#45 (comment)](https://github.com/informalsystems/partnership-heliax/pull/45#discussion_r1174870619) I think for 2., 6. and 7. we're ignoring the start epochs of the src bond deltas, which I think is ok as we never skip a slash on something that should be slashed, only potentially slashing a re-delegation that might not need to be slashed if we were to track the start epochs of re-delegations in detail from the source deltas. I think this is applied consistently across the different paths (unbond/withdraw/end of epoch).

Thanks for the detailed analysis @tzemanovic . I am transforming the spec into a Quint executable spec, so we will be able to test those scenarios much more easily. Regarding your last comment, see my [response](https://github.com/informalsystems/partnership-heliax/pull/45#discussion_r1179039585).

angbrav:
@tzemanovic @cwgoes I've gotten rid of the expensive iterations, in case you want to take a look!

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -636,13 +624,10 @@ end_of_epoch()
...
 
-      forall (val in validators) do
-        var pairs = { pair = <delegator, <epoch, redelegation> | pair in validators[val].redelegations[validator_address] &&
-                                                                 pair !=⊥ && 
-                                                                 redelegation.epoch - unbonding_length <= slash.epoch < redelegation.epoch + pipeline_length}
-        forall (<delegator, redelegation> in pairs) do
-          var staked_redelegated = 
-          slash_redelegated_validator(val, validator_address, slash, redelegation.amount)
+      forall (dest_validators in validators[validator_address].outgoing_redelegations.keys()) do
+        forall (amount in {amount | amount = validators[validator_address].outgoing_redelegations[dest_validators][start] > 0 &&
```
tzemanovic: I think there's a slight difference here from the previous version - the `start` epoch is not defined in this scope. We probably want to iter over the outgoing redelegations for each `validator_address, dest_validators` pair and use the same condition as before for their epochs, i.e. `redelegation.epoch - unbonding_length <= slash.epoch < redelegation.epoch + pipeline_length`

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
tzemanovic: Sry, I'm not sure if I follow. Let me try to ask with 2 examples:

1. Say you have a delegation to validator A with start epoch `e`. In epoch `e+1` you redelegate to validator B. After that, there's evidence discovered for validator A from epoch `e-1`. Would the redelegation get slashed for it when its being processed at `e-1+unbonding` epoch? In this case, the original bond didn't contribute to the validator A voting power when the misbehavior occurred.
2. As another example, let's say that the delegation to validator A has start epoch `e`. Now you redelegate to B in `e+2`, after which an evidence for A is found for `e+1`. Does this differ from the first example when the slash is being processed at `e+1+unbonding` epoch? In this case, the original bond did contribute to the voting power of the validator when it misbehaved.

It seems to me that in both cases the slash would be applied to the redelegation.

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
angbrav: That's the intention of the second condition in the line below: `start - unbonding_length <= slash.epoch < start + pipeline_length`. Maybe it is not well expressed though, we can always make the second iteration more explicit.

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
angbrav: > 1. Say you have a delegation to validator A with start epoch `e`. In epoch `e+1` you redelegate to validator B. After that, there's evidence discovered for validator A from epoch `e-1`. Would the redelegation get slashed for it when its being processed at `e-1+unbonding` epoch? In this case, the original bond didn't contribute to the validator A voting power when the misbehavior occurred.

I see your concern. This is probably a problem: I believe in the current design we would slash. I am not sure there is a way to cut corners and avoid tracking the start of the delegation.

tzemanovic: I think it’s an okay compromise. For redelegation there will be small chance that you may get slashed when you shouldn’t be, but not the other way around so it’s not a safety issue

tzemanovic: Brought this up internally and we agreed we're okay with tracking the re-delegations with start epochs of their respective source bonds to make this precise. We're assuming epoch to last about a day and it's not gonna speed up significantly in future because of other concerns, so this shouldn't pose a DoS vector

angbrav: Addressed in [32102ae](https://github.com/informalsystems/partnership-heliax/pull/45/commits/32102aecfb854459ff15d81bc356898fcbba2d04).

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -1,10 +1,51 @@
...
 ## Assumptions/simplifications
 
 - There is a single token type
 - There is an initial set of active and inactive validators
 - There are always enough non-jalied, candidate validators to fulfill the active set of validators at any given epoch
 
-## Data types
+## Definitions
+
+- `pipeline_length` and `unbonding_length` define two epoch offsets. Note that `pipeline_length` must be greater or equal to `unbonding_length`.
```
cwgoes:
```suggestion
- `pipeline_length` and `unbonding_length` define two epoch offsets. Note that `pipeline_length` must be equal to or greater than `unbonding_length`.
```

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -1,10 +1,51 @@
...
+
+### Redelegations' Specific Terminology
+
+- A redelegation involves two validators. We use `source validator` to refer the validator from which the tokens are taken and `destination validator` to to refer to the validator to which the tokens are delegated.
+
+- Given a `redelegation`, the `redelegation's starting epoch` is the epoch at which the redelegation transaction is processed. The `redelegation's ending epoch` is the epoch at which the redelegated tokens start contributing to the destination delegator. In the current design, if a redelegation's starting epoch is `e`, then its ending epoch would be `e + pipeline_length`. In the text throughout the specification, for a given `redelegation`, we use `redelegation.start` and `redelegation.end` to refer to the starting and ending epoch of the redelegation.
+
+- A redelegation is composed by one or more bonds (aka `redelegation bonds`). Each redelegation bond is a pair the epoch at which the redelegated tokens started contributing to the stake of the source validator and the amount of tokens. It is important to keep track of the starting epochs in order to apply slashes precisely.
```
cwgoes:
```suggestion
- A redelegation is composed of one or more bonds (aka `redelegation bonds`). Each redelegation bond is a pair of the epoch at which the redelegated tokens started contributing to the stake of the source validator and the amount of tokens. It is important to keep track of the starting epochs in order to apply slashes precisely.
```

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -1,10 +1,51 @@
...
+
+- Given a redelegation, we call `redelegation slashing window` the set of consecutive epoch in which the destination validator may be slashed due to the misbehave of the source validator.
```
cwgoes:
```suggestion
- Given a redelegation, we call `redelegation slashing window` the set of consecutive epochs in which the destination validator may be slashed due to the misbehaviour of the source validator.
```

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -1,10 +1,51 @@
...
+  - In the current design, the redelegation slashing window of a redelegation spans from `redelegation.start - unbonding_length` up to `redelegation.end - 1`.
+  - The window's lower bound if determined by the fact that when a redelegation is issued, we apply all slashes of the source validator already processed. This means any slash for an infraction committed at en epoch `< redelegation.start - unbonding_length`.
+  - The window's upper bound is determined by the epoch at which the redelegated tokens stop contributing to the stake to the source validator.
```
cwgoes:
```suggestion
  - The window's upper bound is determined by the epoch at which the redelegated tokens stop contributing to the stake of the source validator.
```

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -222,19 +275,63 @@ tx_withdraw_unbonds_delegator(delegator_address)
 }
 ```
 
-## PoS functions
+The transaction `tx_redelegate` creates a redelegation from source validator `src_validator_address` to `dest_validator_address` for any tokens that a delegator `delegator_address` has currently delegated to the source validator. The transactions takes the following steps:
 
+1. The function first checks that the source validator is not frozen. This is to prevent unbonding at an epoch `e` from a validator that it is known to have behaved, but the slash has not been processed yet.
+2. Then it checks that the redelegation is not a chain redelegation, in which case it prevents it by returning.
+   - The design does not track chain redelegation precisely.
+   - Instead approximates it by preventing a validator to redelegate tokens belonging to a delegator if it has recently received a redelegation of tokens belonging to the same delegator.
```
cwgoes: Recently, i.e. within the unbonding period?

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -222,19 +275,63 @@ tx_withdraw_unbonds_delegator(delegator_address)
...
+   - The `inconming_redelegations` variable is used to implement this mechanism.
```
cwgoes:
```suggestion
   - The `incoming_redelegations` variable is used to implement this mechanism.
```

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -222,19 +275,63 @@ tx_withdraw_unbonds_delegator(delegator_address)
...
+3. The function computes the total amount of tokens that the delegator has bonded at the source validator (`bonded_tokens`).
```
cwgoes, line 287: It should be possible to only redelegate a portion of the tokens that the delegator originally had delegated to the source validator, as well, right?

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -272,66 +371,146 @@ func unbond(validator_address, delegator_address, total_amount)
     //this serves to check that there are bonds (in the docs) and that these are greater than the amount we are trying to unbond
     if (sum{amount | <start, amount> in delbonds} >= total_amount) then
       var remain = total_amount
-      var amount_after_slashing = total_amount
-      //Iterate over bonds and create unbond
+      var amount_after_slashing = 0
+      //iterate over bonds and create unbonds
       forall (<start, amount> in delbonds while remain > 0) do
+        //Retrieve the bond's associated redelegation record (if any)
```
cwgoes:
```suggestion
        //Retrieve the bond's associated redelegation records (if any)
```

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -357,6 +536,182 @@ func new_evidence(evidence)
 }
 ```
 
+```go
+func compute_stake_fraction(infraction, voting_power){
+  switch infraction
+    case duplicate_vote: return duplicate_vote_rate * voting_power
+    case ligth_client_attack: return ligth_client_attack_rate * voting_power
```
cwgoes:
```suggestion
    case light_client_attack: return light_client_attack_rate * voting_power
```

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
```diff
@@ -357,6 +536,182 @@ func new_evidence(evidence)
...
+    default: panic()
+}
+```
+
+The function `end_of_epoch` is called at the end of every epoch. It takes the following steps:
+
+1. It first computes the slash rate for any slash that has to be processed at the end of the current epoch.
+2. The iterates over the enqueued slashes. For each slash:
```
cwgoes:
```suggestion
2. It then iterates over the enqueued slashes. For each slash:
```

cwgoes (commented):
Broadly this seems alright. I agree that the model ended up being quite complex/intricate, easy to make a mistake here. Hope that Quint will help, and I think we should also try to write down the properties that we expect to be preserved w.r.t. conservation-of-tokens, slashing, etc. and write proptests for those in our Rust codebase.

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
angbrav: Sort of. I have added some text to clarify, but the text in definitions about chain redelegation explains this better.

2023/Q1/artifacts/PoS-pseudocode/PoS-model-redelegation.md
angbrav: It should be possible. Do we want to do it? I could look into it.

cwgoes: Yes, this would be nice (at some point, it's not urgent)

cwgoes:
@angbrav should we merge this?

angbrav:
We could, but it is not in sync with the Quint spec. Do we want to spend some time correcting it?

angbrav:
Closed as agreed to stop maintaining pseudocode specs in favor of Quint specs.

PR #50, opened 2023-05-10
PoS updates to ensure non-negative stakes + misc upgrades

brentstone:
Based on #43.

The changes in this PR were initiated by the need to adjust the pseudocode to reflect changes in the Namada implementation that prevent the validator's deltas from going negative with the existence of more arbitrary slashing conditions.

Some other pseudocode updates that are not directly related to this are included as well.

### DIFF
--- a/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
+++ b/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
@@ -271,7 +271,8 @@ func unbond(validator_address, delegator_address, total_amount)
     var delbonds = {<start, amount> | amount = bonds[delegator_address][validator_address].deltas[start] > 0 && start <= cur_epoch + pipeline_len}
     if (sum{amount | <start, amount> in delbonds} >= total_amount) then
       var remain = total_amount
-      var amount_after_slashing = 0
+      var amount_after_slashing = 0 // Track the amount to change the deltas (may be less than `total_amount` due to previous slash processing)
+
       // Iterate over bonds and create unbond
       forall (<start, amount> in delbonds while remain > 0) do
         // Get the minimum of the remainder and the unbond, equal to amount if remain > amount and remain otherwise 
--- a/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
+++ b/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
@@ -74,6 +74,7 @@ type ValidatorSet struct {
 ```go
 pipeline_length uint
 unbonding_length uint
+cubic_slash_window_width uint
 min_sentence uint
 votes_per_token uint
 duplicate_vote_rate float
@@ -274,13 +275,14 @@ func unbond(validator_address, delegator_address, total_amount)
     if (sum{amount | <start, amount> in delbonds} >= total_amount) then
       var remain = total_amount // Track the remaining bond amount to unbond
       var amount_after_slashing = 0 // Track the amount to change the deltas (may be less than `total_amount` due to previous slash processing)
+      var withdraw_epoch = cur_epoch + pipeline_length + unbonding_length + cubic_slash_window_width
 
       // Iterate over bonds and create unbond
       forall (<start, amount> in delbonds while remain > 0) do
         // Get the minimum of the remainder and the unbond, equal to amount if remain > amount and remain otherwise 
         var amount_unbonded = min{amount, remain}
         bonds[delegator_address][validator_address].deltas[start] = amount - amount_unbonded
-        unbonds[delegator_address][validator_address].deltas[start, cur_epoch+pipeline_length+unbonding_length] += amount_unbonded
+        unbonds[delegator_address][validator_address].deltas[start, withdraw_epoch] += amount_unbonded
         // Set of slashes that happened while the bond was contributing to the validator's stake
         var set_slashes = {s | s in slashes[validator_address] && start <= slash.epoch }
         amount_after_slashing += compute_amount_after_slashing(set_slashes, amount_unbonded)
@@ -308,8 +310,8 @@ func withdraw(validator_address, delegator_address)
 {
   var delunbonds = {<start,end,amount> | amount = unbonds[delegator_address][validator_address].deltas[(start, end)] > 0 && end <= cur_epoch }
   forall (<start,end,amount> in selfunbonds) do
-    var set_slashes = {s | s in slashes[validator_address] && start <= slash.epoch && slash.epoch < end - unbonding_length }
     // Apply slashes that happened while the bond associated with the unbond was contributing to the validator's stake
+    var set_slashes = {s | s in slashes[validator_address] && start <= slash.epoch && slash.epoch < end - unbonding_length - cubic_slash_window_width}
     var amount_after_slashing = compute_amount_after_slashing(set_slashes, amount)
     balance[delegator_address] += amount_after_slashing
     balance[pos] -= amount_after_slashing
@@ -350,8 +352,8 @@ func new_evidence(validator, infraction_epoch, type)
   // Create slash
   var total_staked = read_epoched_field(validators[evidence.validator].total_deltas, infraction_epoch, 0)
   var slash = Slash{epoch: infraction_epoch, validator: validator, rate: 0, type: type, voting_power: total_staked}
-  append(enqueued_slashes[infraction_epoch + unbonding_length], slash)
   // Enqueue slash (Step 1.1 of cubic slashing)
+  append(enqueued_slashes[infraction_epoch + unbonding_length + cubic_slash_window_width], slash)
   // Jail validator (Step 1.2 of cubic slashing)
   validators[validator].jail_record = JailRecord{is_jailed: true, epoch: cur_epoch+1}
   remove_validator_from_sets(validator, 1)
@@ -508,7 +510,7 @@ func get_min_slash_rate(infraction){
 end_of_epoch()
 {
 // The infraction epoch is the same for all enqueued slashes (`slash.epoch` in below code)
-  var infraction_epoch = cur_epoch - unbonding_length
+  var infraction_epoch = cur_epoch - unbonding_length - cubic_slash_window_width
   // Iterate over all slashes for infractions within (-1,+1) epochs range (Step 2.1 of cubic slashing)
   var set_slashes = {s | s in enqueued_slashes[epoch] && cur_epoch-1 <= epoch <= cur_epoch+1}
   // Calculate the cubic slash rate for all slashes processed this epoch (Step 2.2 of cubic slashing)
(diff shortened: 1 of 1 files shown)


2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -301,34 +303,43 @@ func unbond(validator_address, delegator_address, total_amount)
...
-        updated_amount -= slashed_amount.amount
-        computed_amounts = computed_amounts \ {slashed_amount}
-      computed_amounts = computed_amounts \union {SlashedAmount{epoch: slash.epoch, amount: updated_amount*slash.rate}}
+  var updated_amount = amount
 
-    return updated_amount - sum({computed_amount.amount | computed_amount in computed_amounts})
+  // Now apply the slashes
+  forall (slash in slash_rates in slash.epoch order) do
+    // Update amount with slashes that happened more than `unbonding_length` before this slash
+    forall (slashed_amount in computed_amounts s.t. slashed_amount.epoch + unbonding_length < slash.epoch) do
```
brentstone: Should we now add the window width here: `s.t. slashed_amount.epoch + unbonding_length + cubic_slash_window_width < slash.epoch`?

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -501,46 +512,82 @@ func compute_stake_fraction(infraction, voting_power){
...
-        update_voting_power(validator_address, offset)
-        total_unbonded = 0
+    // Find the total amount deducted from the deltas due to unbonds that became active after the infraction epoch. This is used to ensure the deltas are appropriately slashed.
+    // Note: need to only do this once
+    var total_unbonded = 0
+    // Find the total unbonded from the slash epoch up to the current epoch first
+    // a..b notation determines an integer range: all integers between a and b inclusive
+    forall (epoch in slash.epoch+1..cur_epoch) do
+      forall ((unbond_start, unbond_amount) in validators[validator_address].set_unbonds[epoch] s.t. unbond_start <= infraction_epoch)
+        var set_prev_slashes = {s | s in slashes[validator_address] && unbond_start <= s.epoch && s.epoch + unbonding_length < infraction_epoch}
```
brentstone: Should we now add the window width here: `s.epoch + unbonding_length + cubic_slash_window_width < infraction_epoch`?

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -501,46 +512,82 @@ func compute_stake_fraction(infraction, voting_power){
...
+    // Up to pipeline_length because there cannot be any unbond in a greater epoch (cur_epoch+pipeline_length is the upper bound)
+    forall (offset in 1..pipeline_length) do
+      forall ((unbond_start, unbond_amount) in validators[validator_address].set_unbonds[cur_epoch + offset] s.t. unbond_start <= infraction_epoch) do
+        // We only need to apply a slash s if s.epoch < unbond.end - unbonding_length
+        // It is easy to see that s.epoch + unbonding_length < slash.epoch implies s.epoch < unbond.end - unbonding_length
+        // 1) slash.epoch = cur_epoch - unbonding_length
+        // 2) unbond.end = cur_epoch + offset + unbonding_length => cur_epoch = unbond.end - offset - unbonding_length
+        // By 1) s.epoch + unbonding_length < cur_epoch - unbonding_length
+        // By 2) s.epoch + unbonding_length < unbond.end - offset - 2*unbonding_length => s.epoch < unbond.end - offset - 3*unbonding_length, as required.
+        var set_prev_slashes = {s | s in slashes[validator_address] && unbond_start <= s.epoch && s.epoch + unbonding_length < infraction_epoch}
```
brentstone: Should we now add the window width here: `s.epoch + unbonding_length + cubic_slash_window_width < infraction_epoch`?

angbrav:
I am adding these changes to the Quint spec to see what happens. I'll get then back to you :)

angbrav:
> Some other pseudocode updates that are not directly related to this are included as well.

Can we list them here? It will help me review it.

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -301,34 +304,43 @@ func unbond(validator_address, delegator_address, total_amount)
...
   forall (<start,end,amount> in selfunbonds) do
-    // set of slashes that happened while the bond associated with the unbond was contributing to the validator's stake
-    var set_slashes = {s | s in slashes[validator_address] && start <= slash.epoch && slash.epoch < end - unbonding_length }
+    // Apply slashes that happened while the bond associated with the unbond was contributing to the validator's stake
+    var set_slashes = {s | s in slashes[validator_address] && start <= slash.epoch && slash.epoch < end - unbonding_length - cubic_slash_window_width}
     var amount_after_slashing = compute_amount_after_slashing(set_slashes, amount)
     balance[delegator_address] += amount_after_slashing
-    balance[pos] -= amount_after_slashing
-    //remove unbond
+    balance[slash_pool] += amount - amount_after_slashing
```
angbrav, line 317: I believe we should only transfer tokens to the `slash_pool` address at the end of an epoch. The tokens being transferred here have already been transferred at the end of the epoch when the slash was processed.

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
angbrav: I think so

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
angbrav: I think so

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
angbrav: I think so

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -492,56 +504,81 @@ func is_validator(validator_address, epoch){
...
-        update_total_deltas(validator_address, offset, diff_slashed_amount)
-        update_voting_power(validator_address, offset)
-        total_unbonded = 0
+    // Find the total amount deducted from the deltas due to unbonds that became active after the infraction epoch. This is used to ensure the deltas are appropriately slashed.
+    // Note: need to only do this once
+    var total_unbonded = 0
+    // Find the total unbonded from the slash epoch up to the current epoch first
+    // a..b notation determines an integer range: all integers between a and b inclusive
+    forall (epoch in slash.epoch+1..cur_epoch) do
+      forall ((unbond_start, unbond_amount) in validators[validator_address].set_unbonds[epoch] s.t. unbond_start <= infraction_epoch)
```
angbrav: This may be redundant, but just in case to help readability: it is unclear whether the inner map (Epoch -> int) is initialized for all epochs with 0 amount or not. The Quint spec makes this explicit and precise.

```suggestion
      forall ((unbond_start, unbond_amount) in validators[validator_address].set_unbonds[epoch] s.t. unbond_start <= infraction_epoch && unbond_amount > 0)
```

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -492,56 +504,81 @@ func is_validator(validator_address, epoch){
...
+        var set_prev_slashes = {s | s in slashes[validator_address] && unbond_start <= s.epoch && s.epoch + unbonding_length < infraction_epoch}
+        total_unbonded += compute_amount_after_slashing(set_prev_slashes, unbond_amount)
+
+    // For the future epochs, do the same as before but also update the deltas
+    var total_slashed = 0
+    var last_slash = 0
+    // Up to pipeline_length because there cannot be any unbond in a greater epoch (cur_epoch+pipeline_length is the upper bound)
+    forall (offset in 1..pipeline_length) do
+      forall ((unbond_start, unbond_amount) in validators[validator_address].set_unbonds[cur_epoch + offset] s.t. unbond_start <= infraction_epoch) do
```
angbrav: This may be redundant, but just in case to help readability: it is unclear whether the inner map (Epoch -> int) is initialized for all epochs with 0 amount or not. The Quint spec makes this explicit and precise.
```suggestion
      forall ((unbond_start, unbond_amount) in validators[validator_address].set_unbonds[cur_epoch + offset] s.t. unbond_start <= infraction_epoch && unbond_amount > 0) do
```

angbrav (commented):
- I've reviewed the total_unbonded optimization and looks good to me. It is already in the Quint spec and seems to be working fine. Can we maybe change the variable name `set_unbonds` to something else (`total_unbonded` for instance)? it is not a set anymore.
- Adding `cubic_slash_window_width` seems fine. I am adding to Quint to see what happens.
- I have not checked the remaining: (i) preventing stake becoming zero; (ii) slash pool. Although I have left a comment about the latter.

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
angbrav: I think so for all the cases.

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
brentstone: Not all of the tokens transferred here have been transferred during `end_of_epoch` I believe. During `end_of_epoch`, the token amount transferred is the additional amount of tokens to deduct from the deltas such that we achieve the desired total slash amount when considering already existing unbonds of bonds that contributed to the infraction.

The amount transferred during the withdrawal here is the amount of these already existing unbonds (which already reduced the deltas before processing the slashes) that need to be sent to the slash pool. Does this make sense?

angbrav: I am not sure I followed your text, but let me try to elaborate a bit more my point. On withdrawing, we are going to apply slashes that have been already processed. This means that we have already taken from `total_deltas` the slashable amount and moved it to the slash pool. We simply do slashing at withdrawing because we do not update bonds or unbonds when slashing at the end of an epoch, so we need to apply the (already proccessed) slashes to those unbonds in order to transfer the appropriate amount to users (otherwise we would be transferring too many tokens in some cases)

I would say that if this is not the case, then users should be able to withdraw tokens before concerning (those affecting the withdrawn unbonds) slashes are processed, even when the evidence is submitted on time.

angbrav:
> * I've reviewed the total_unbonded optimization and looks good to me. It is already in the Quint spec and seems to be working fine. Can we maybe change the variable name `set_unbonds` to something else (`total_unbonded` for instance)? it is not a set anymore.
> * Adding `cubic_slash_window_width` seems fine. I am adding to Quint to see what happens.

Done, looks good :)

> * I have not checked the remaining: (i) preventing stake becoming zero; (ii) slash pool. Although I have left a comment about the latter.

@brentstone what do you think about splitting once more this PR into two: one with the optimization for `total_unbonded` and  `cubic_slash_window_width`, which I believe would be ready to merge; and a second one with the logic specific for preventing stake becoming zero and the slash pool?

brentstone:
Closed in favor of splitting out into #54 and #55.

PR #55, opened 2023-05-27
Prevent overslashing and negative stakes

### LINKED ISSUE #46: Fix validator over-slashing from multiple infractions in the same epoch
If a validator commits more than one infraction in the same epoch, the current pseudocode can over-slash and sometimes even cause the validator stake to go negative. We should handle multiple slashes from one validator by adding their rates for an effective slash rate that is capped at 100%.

brentstone:
Closes #46.

The changes in this PR were initiated by the need to adjust the pseudocode to reflect changes in the Namada implementation that prevent the validator's deltas from going negative with the existence of more arbitrary slashing conditions.

### DIFF
--- a/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
+++ b/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
@@ -32,6 +32,7 @@ type Validator struct {
   state map<Epoch, {inactive, candidate}>
   total_deltas map<Epoch, amount:int>
   unbond_records map<Epoch, map<Epoch, amount:int>> // outer epoch for the unbond start (when the tokens stopped contributing to the validator's stake), inner for the underlying bond start
+  total_bonded map<Epoch, amount:int>
   voting_power map<Epoch, VotingPower>
   reward_address Addr
   jail_record JailRecord
@@ -239,6 +240,7 @@ func bond(validator_address, delegator_address, amount)
   if is_validator(validator_address, cur_epoch+pipeline_length) then
     // Add bond amount to deltas at n + pipeline_length
     bonds[delegator_address][validator_address].deltas[cur_epoch+pipeline_length] += amount
+    validators[validator_address].total_bonded[cur_epoch+pipeline_length] += amount
     // Debit amount from delegator account and credit it to the PoS account
     balances[delegator_address] -= amount
     balances[pos] += amount
@@ -536,12 +538,20 @@ end_of_epoch()
     // Find the total amount deducted from the deltas due to unbonds that became active after the infraction epoch. This is used to ensure the deltas are appropriately slashed.
     // Note: need to only do this once
     var total_unbonded = 0
+    var sum_post_bonds = 0
+
     // Find the total unbonded from the slash epoch up to the current epoch first
     // a..b notation determines an integer range: all integers between a and b inclusive
     forall (epoch in slash.epoch+1..cur_epoch) do
       forall ((unbond_start, unbond_amount) in validators[validator_address].unbond_records[epoch] s.t. unbond_start <= infraction_epoch && unbond_amount > 0)
         var set_prev_slashes = {s | s in slashes[validator_address] && unbond_start <= s.epoch && s.epoch + unbonding_length + cubic_slash_window_width < infraction_epoch}
         total_unbonded += compute_amount_after_slashing(set_prev_slashes, unbond_amount)
+
+      // Do not over-slash or improperly slash stake that did not exist at the infraction epoch
+      var recent_unbonds = 0
+      forall ((bond_start, amount) in validators[validator_address].unbond_records[epoch] s.t. amount > 0 && bond_start > infraction_epoch) do
+        recent_unbonds += amount  
+      sum_post_bonds += validators[validator_address].total_bonded[epoch] - recent_unbonds
 
     // For the future epochs, do the same as before but also update the deltas
     var last_slash = 0
@@ -561,7 +571,11 @@ end_of_epoch()
       last_slash = this_slash
 
       // Do not over-slash or improperly slash stake that did not exist at the infraction epoch
-      var sum_post_bonds = sum{bonds[validator_address].deltas[start] s.t. start > infraction_epoch && start <= cur_epoch + offset>}
+      var recent_unbonds = 0
+      forall ((bond_start, amount) in validators[validator_address].unbond_records[cur_epoch+offset] s.t. amount > 0 && bond_start > infraction_epoch) do
+        recent_unbonds += amount  
+      sum_post_bonds += validators[validator_address].total_bonded[cur_epoch+offset] - recent_unbonds
+      
       var validator_stake = read_epoched_field(validators[validator_address].total_deltas, cur_epoch + offset, 0)
       var slashable_stake = validator_stake - sum_post_bonds
       var change = min{-slashable_stake, diff_slashed_amount}
--- a/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
+++ b/2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
@@ -508,6 +508,7 @@ func get_min_slash_rate(infraction){
 // Processes the enqueued slashes by calculating the cubic slashing rate and then slashing the validator's deltas (stake)
 end_of_epoch()
 {
+  // The infraction epoch is the same for all enqueued slashes
   var infraction_epoch = cur_epoch - unbonding_length - cubic_slash_window_width
   // Iterate over all slashes for infractions within (- window, + window) epochs range (Step 2.1 of cubic slashing)
   var set_slashes = {s | s in enqueued_slashes[epoch] && cur_epoch - cubic_slash_window_width <= epoch <= cur_epoch + cubic_slash_window_width}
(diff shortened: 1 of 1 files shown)


2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -316,16 +322,24 @@ func withdraw(validator_address, delegator_address)
 
 ```go
 compute_amount_after_slashing(set_slashes, amount) {
+  // First, group the slashes by epoch and a total rate
```
angbrav, line 322: Why do we need to do this?

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -289,7 +289,13 @@ func unbond(validator_address, delegator_address, total_amount)
 
         validators[validator_address].unbond_records[cur_epoch+pipeline_length][start] += amount_unbonded
         remain -= amount_unbonded
-      update_total_deltas(validator_address, pipeline_length, -1*amount_after_slashing)
+      
+      // Ensure that the validator's stake does not go negative due to the slashing
+      var pipeline_stake = read_epoched_field(validators[validator_address].total_deltas, cur_epoch + pipeline_length, 0)
+      var token_change = min{ amount_after_slashing, pipeline_stake }
```
angbrav: Can you think in a scenario in which this is needed? Intuitively I'd say we do not need it. We are computing how much from the bond is left after slashing and we are concerned about overslashing because the validator misbehaved at least twice with the same tokens. In those cases, `amount_after_slashing` would be 0. But I do not know if I am missing some scenarios in which the validator is slashed all its stake at the end of an epoch, and then when applying the slashes to individual bonds here some of the bonds won't be fully slashed... that would be a case in which we would need to be careful at this point...

angbrav (commented):
The only thing that's unclear to me is the comment about capping in unbonding. I am integrating these changes to the Quint spec to see what happens. BTW, what about the slash pool? I think it did not make it into any PR.. it is fine like that, but we should not forget about it.

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
brentstone: Slashes within the same epoch will always be applied by summing their rates, so we can essntially turn all slashes into one effective slash per epoch. These lines also cap the slash rate per epoch at 1.0

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
brentstone: I think you may be right in thinking this may not be necessary. It's possible I threw it in there when I was initially trying to prevent negative stake in lots of places. I've just tested the removal of this logic with our native state machine test and the test passed.

I'll remove this for now and remove it from the Namada code in a new PR.

2022/Q4/artifacts/PoS-pseudocode/PoS-model.md
```diff
@@ -509,37 +520,54 @@ end_of_epoch()
...
+        total_unbonded += compute_amount_after_slashing(set_prev_slashes, unbond_amount)
+      var this_slash = (total_staked - total_unbonded) * total_rate
+      var diff_slashed_amount = last_slash - this_slash
+      last_slash = this_slash
+
+      // Do not over-slash or improperly slash stake that did not exist at the infraction epoch
+      var sum_post_bonds = sum{bonds[validator_address].deltas[start] s.t. start > infraction_epoch && start <= cur_epoch + offset>}
+      var validator_stake = read_epoched_field(validators[validator_address].total_deltas, cur_epoch + offset, 0)
+      var slashable_stake = validator_stake - sum_post_bonds
+      var change = min{-slashable_stake, diff_slashed_amount}
```
angbrav: This part I am not sure about. I think we just need to compute how much from this_slash is left after applying any preceding slash. Could not figure out if what it is in here is equivalent, but we should avoid iterating over bonds at the end of an epoch.

angbrav: I keep thinking about this. I believe that to implement this correctly we need to be able to answer the following question: given an epoch `e` and a validator `val`, how much stake is left at an epoch `e' > e` from the stake the validator had at `e`. I haven't been able to figure out how to answer this question with the state variables that we have and without iterating over bonds/unbonds.

My guess is that your code tries to do something like that, but I am not sure it succeeds in doing so. My concern is related to unbonds that remove bonds but are only effective at the `pipeline_length` offset.

angbrav: I am proposing the following: keep a new variable `total_bonded` for each validator which is a map from Epoch to int. 
We update the variable when bonding as follows:
`validators[validator_address].total_bonded[cur_epoch+pipeline_length] += amount`

Given a `offset`, we use this state variable as follows to substitute lines 564.
```go
var recent_unbonds = 0
forall (let [bond_start, amount] of validators[validator_address].unbond_records[cur_epoch+offset] s.t. amount > 0 &&
                                                                      bond_start > infraction_epoch) do
  recent_unbonds += amount                                                                    
sum_post_bonds = sum_post_bonds + (validators[validator_address].total_bonded[cur_epoch+offset] - recent_unbonds)

```
Note that `sum_post_bonds` is always `>= 0` (we are only interested in the bonds that were added after the infraction epoch and we subtract how much of those have been unbonded) and that `sum_post_bonds` is computed iteratively within the for all loops. How does this look for you? It is similar to your code but it addresses to issues:
- Does not iterate over bonds
- I am not 100% sure, but I think your code has the issue of not accounting for unbonds properly.

brentstone: Does the commit 01a5028596df73c690401f783840c902183b3bd0 satisfy everything you had in mind here?

angbrav: I think we should start computing `sum_post_bonds` in the previous `forall`: the one that goes from `slash.epoch+1` to `cur_epoch`. Otherwise, we would be possibly missing a bunch of bonds.

brentstone: fixed in 16d4268

brentstone: Have you thought of an example where the previous code fails to account for things but your proposed code successfully does? I haven't been able to do so yet.

In your code, `total_bonded` is not decremented when unbonding, but is rather effectively done so when we are computing slashes with `recent_unbonds`. On the contrary, the original code seems to do this more eagerly when unbonding rather. I'm still wondering about a difference between doing it while slashing vs. while unbonding.

angbrav: The difference is that `recent_unbonds` only counts those tokens that were bonded after the infraction epoch and the way your proposal would compute `total_bonded` does not make that selection. Let me try to sketch a scenario:
- Validator misbehaves at `e` but the infraction is discovered at `e+unbonding_length`. This means users can unbond from the validator from `e` to `e+unbonding_length` (until the infraction is discovered).
- Assume that a user unbonds `x` tokens that were bonded at an epoch `e' < e` at epoch `e+1`. in your proposal, we would have `total_bonded[e+1]=-x`.
-  This will result in `sum_post_bonds` being negative when the infraction is processed at the end of `e+unbonding_length+cubic_offset`, which, if I am not mistaken, is a problem.

brentstone: Ah, so this is not something that would occur in our current implementation. Rather, when unbonding, the change would be `total_bonded[e'] -= x`; `total_bonded` works exactly the same as `bonds[delegator_address][validator_address]`, except folds in all of the delegators such that it is `total_bonded[validator_address]`. Thus, `total_bonded` would never go negative as long as `bonds` cannot go negative.

I'll try to point out some relevant code in Namada that shows what we are doing. Hope this can possibly be helpful, though I understand you probably haven't seen any of our Rust code before.

In Namada, we are using something called [`global_bond_handle`](https://github.com/anoma/namada/blob/main/proof_of_stake/src/lib.rs#L304-L308) to hold the sum of all bonds to a validator in the same way as the bonds that are prefixed by delegator and validator address. This is the object analagous to `total_bonded`.

When we bond tokens, we update both the normal bonds and the global bonds in the same way ([see code here](https://github.com/anoma/namada/blob/main/proof_of_stake/src/lib.rs#L911-L925)).

When we unbond tokens, we decrement the bond values in place for both the normal bonds and the global bonds. As long as the normal bond has a nonzero value, the global bond is also guaranteed to have nonzero value. See the relevant code [here](https://github.com/anoma/namada/blob/main/proof_of_stake/src/lib.rs#L1612-L1659), particularly [these lines](https://github.com/anoma/namada/blob/main/proof_of_stake/src/lib.rs#L1629-L1630).

brentstone: Tried to put this in pseudocode quickly here: ac99310bed235cbb5c8806ec44077a62bb975994

angbrav: > Tried to put this in pseudocode quickly here: https://github.com/informalsystems/partnership-heliax/commit/ac99310bed235cbb5c8806ec44077a62bb975994

That may not work. Let's say you bond `x` tokens at epoch `e` and unbond those tokens at epoch `e+2` and you are computing `sum_post_bonds` at epoch `e`. Then the stake at `e` includes those `x` tokens but your `global_bonds[validator_address].deltas[start]` won't: you subtract them when unbonding. This looks incorrect to me.

brentstone: Sorry, I think I need more context for your example here... if I'm computing `sum_post_bonds` at epoch `e` in my version of things, then there is no way that an unbond of tokens can remove stake at `e+2` because the validator will be frozen already by epoch `e` when the unbond tx is submitted (only computing `sum_post_bonds` for future epochs relative to the current).

angbrav: - Assume that `unbonding_length=4`, `pipeline_length=3` and `cubic_offset=1`
- A validator misbehaves at `e`. The infraction is submitted at `e+unbonding_length=e+4` and processed at the end of `e+unbonding_length+cubic_offset=e+5`
- A user delegates `x` tokens to the validator at `e+1` and unbonds them at `e+unbonding_length=e+4` (before the infraction is submitted). 
- When processing the infraction at the end of `e+5`, we would compute `sum_post_bonds` for `e+6` (`cur_epoch+1`). By that epoch, the unbond has not materialized (it materializes at `e+4+pipeline_length=e+7)`, but `global_bonds[e+1]` accounts for it.

I know those are not typical values for `unbonding_length` and `pipeline_length`, but I'd rather not have the correctness of the design depend on the value of these parameters.

brentstone: Thx for this example. I agree that your proposed changes do indeed solve a problem that currently exists. I wrote up a comparison of the two approaches that demonstrates where the current one fails and where the proposed one succeeds: https://hackmd.io/yRsLbn5IT8C2xkj8csSZMg?both

angbrav:
Looks good, merged

PR #54, opened 2023-05-26
Add cubic slash window width and refactor `set_unbonds`

brentstone:
The two primary changes are:
- add the cubic slash window width to various offsets for processing and collecting slashes, determing when unbonds can be withdrawn
- Refactor and rename `set_unbonds`

Additionally, some documentation is improved and some minor fixes brought up in other github discussions are included.

PR #56, opened 2023-05-27
Parameterize the eligible slashes for cubic slashes by the window

brentstone:
Forgot to include this as part of #54.


angbrav (approved):
Thanks!

PR #58, opened 2023-06-12
Quint spec review

brentstone:
Just some miscellaneous fixes and items for review as I familiarize myself with the code.

### DIFF
--- a/2023/Q2/artifacts/PoS-quint/namada.qnt
+++ b/2023/Q2/artifacts/PoS-quint/namada.qnt
@@ -33,7 +33,7 @@ module namada {
     type Unbond = {
       // Epoch at which the unbond tokens started contributing to the validator's stake
       start: Epoch,
-      // Epoch at which the unbonded toeksn can be withdrawn
+      // Epoch at which the unbonded tokens can be withdrawn
       end: Epoch,
       // Amount of unbonded tokens
       amount: int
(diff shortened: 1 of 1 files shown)


brentstone:
@angbrav @tzemanovic each commit here leaves a comment or fixes smth that I think should be changed

2023/Q2/artifacts/PoS-quint/namada.qnt
```diff
@@ -187,7 +187,7 @@ module namada {
       if (amount <= delegatorState.balance and amount > 0) {
         val updatedDelegatorState = delegatorState.with("balance", delegatorState.balance - amount)
                                                   .with("bonded", delegatorState.bonded.set(validatorState.address,
-                                                                                            delegatorState.bonded.get(validatorState.address).setAdd({start: posState.epoch + PIPELINE_OFFSET, amount: amount})))                                                                       
+                                                                                            delegatorState.bonded.get(validatorState.address).setAdd({start: posState.epoch + PIPELINE_OFFSET, amount: amount}))) // is this always adding a new bond element rather than updating? maybe this doesn't matter if only 1 tx is permitted, but from a spec perspective not sure if this is incorrect.
```
tzemanovic, line 190: yeah, I think this is related to the comment on:

```quint
    // Transactions per epoch
    // the spec is not fully ready to handle multiple txs per epoch
    // we would need to handle duplicate bonds and unbonds
    pure val TXS_EPOCH: int = 1
```

I think we could probably change `bonded: Address -> Set[Bond]` to `bonded: Address -> Epoch -> int` to handle that

2023/Q2/artifacts/PoS-quint/namada.qnt
```diff
@@ -493,7 +493,7 @@ module namada {
         pos' = result.posState,
         delegators' = delegators,
         lastTx' = {tag: "Evidence", result: true, user: "", validator: validator, amount: e},
-        nextInfractionEpoch' = nextInfractionEpoch.set(validator, e + UNBONDING_OFFSET)
+        nextInfractionEpoch' = nextInfractionEpoch.set(validator, e + UNBONDING_OFFSET) // should this have CUBIC_OFFSET?
```
tzemanovic, line 496: I think so to correspond to the offset of `enqueuedSlashes`

2023/Q2/artifacts/PoS-quint/namada.qnt
angbrav: Exactly, https://github.com/informalsystems/partnership-heliax/pull/63 addresses it.

2023/Q2/artifacts/PoS-quint/namada.qnt
angbrav: This is only to control the pace at which infractions of a given validator can be submitted without preventing repeated infractions: a validator misbehaving more than once with the same stake. To prevent repeated infractions, we can set `limitEvidence=2`

PR #79, opened 2023-09-22
Fix slash validator

### LINKED ISSUE #74: Overslashing because of the way negative stakes are prevented
@tzemanovic found a problem with the current design. The problem is the way we prevent negative stakes due to a validator misbehaving more than once with the same stake. In same cases, we may overslash a delegator. See the following example:
- Assume that a validator starts with `X` tokens (self-bond) and then receives a delegation of `Y` tokens at an epoch `e_1, i.e., contributing to the validator's stake starting from e_1+pipeline_offset.
- Assume that the validator misbehaves twice: once at `e_1` and a second one at an epoch `e_2>=e_1+pipeline_offset`. Thus, the first infraction affects the initial `X` self-bonded tokens, and the second both.
- Assume that the slash rate for both is `0.7` (to illustrate the issue in this execution, we only need the slash rate of the first infraction to be greater than `0.5`)
- When processing the first slash, we slash `0.7*X`.
- When processing the second infraction, we should slash `0.7*X + 0.7*Y`. Since there are no enough tokens left from the self-bond (only `0.3*X` left), we will take some from the delegation. In the worst case, all that are left (`0.3*Y`).
- This is because we compute `slashableAmount` by summing all that is left from validator's stake at the infraction epoch.

angbrav:
This PR fixes #74. It includes the following changes:
- New approach for slashing a misbehaving validator, i.e., new `slashValidator` function
- Adds a bunch of helper functions and removes some that are not used anymore (used by the old `slashValidator` function
- Adds an invariant to check that the validator's voting power at a given epoch is equal to the total amount of tokens delegated to that validator after applying slashes
- Adds tests for all new functions

I've run the Quint simulator and no violation was found. I've run the following experiments:

40 steps, 100k, allInvariantsWithNoSlashPool, Unbonding=4, pipeline=2, limitEvidence=2 (~80minutes)
40 steps, 100k, allInvariantsWithNoSlashPool, Unbonding=4, pipeline=2, limitEvidence=1 (~80minutes)
40 steps, 100k, allInvariantsWithNoSlashPool, Unbonding=4, pipeline=2, limitEvidence=0 (~80minutes)

### DIFF
--- a/2023/Q3/artifacts/PoS-quint/helper-functions.qnt
+++ b/2023/Q3/artifacts/PoS-quint/helper-functions.qnt
@@ -875,6 +875,79 @@ module helperFunctions {
       result._4
     }
 
+    // The slashBond function computes how much is left from a bond after applying a set of slashes.
+    // - @param start the epoch at which the bond started.
+    // - @param amount the bonded amount.
+    // - @param redelegatedBond a RedelegatedBondsMap that .
+    // - @param validator the address of the bonding validator.
+    // - @param slashes a map from address validator to a list of slashes.
+    // - @returns the amount of tokens left after slashing.
+    pure def computeBondAtEpoch(epoch: Epoch, start: Epoch, amount: int, validator: Address, slashes: Address -> List[Slash], redelegatedBond: RedelegatedBondsMap): int = {
+       val listSlashes = slashes.get(validator).select(s => epoch <= s.epoch and s.epoch + slashProcessingDelay < epoch)
+       val filteredSlashMap = slashes.keys().mapBy(v => slashes.get(v).select(s => s.epoch + slashProcessingDelay < epoch))
+       val resultFold = foldAndSlashRedelegatedBondsMap(redelegatedBond, start, listSlashes, filteredSlashMap)
+       val totalNoRedelegated = amount - resultFold.totalRedelegated
+       val afterNoRedelegated = applyListSlashes(listSlashes, totalNoRedelegated)
+       afterNoRedelegated + resultFold.totalAfterSlashing
+    }
+
+
+    /*pure def slashBond(start: Epoch, amount: int, redelegatedBond: RedelegatedBondsMap, validator: Address, slashes: Address -> List[Slash]): int = {
+      val listSlashes = slashes.get(validator).select(slash => start <= slash.epoch)
+      val resultFold = foldAndSlashRedelegatedBondsMap(redelegatedBond, start, listSlashes, slashes)
+      val totalNoRedelegated = amount - resultFold.totalRedelegated
+      val afterNoRedelegated = applyListSlashes(listSlashes, totalNoRedelegated)
+      afterNoRedelegated + resultFold.totalAfterSlashing
+    }
+
+    run slashBondTest = {
+      val slashes = VALIDATORS.mapBy(v => if (v == "alice") [{epoch: 4, rate: 1}] else List())
+      all {
+        assert(slashBond(10, 30,  Map(), "alice", VALIDATORS.mapBy(v => List())) == 30),
+        assert(slashBond(10, 30,  Map(), "alice", VALIDATORS.mapBy(v => if (v == "alice") [{epoch: 4, rate: 1}] else List())) == 30),
+        assert(slashBond(10, 30,  Map(), "alice", VALIDATORS.mapBy(v => if (v == "alice") [{epoch: 11, rate: 1}] else List())) == 0),
+        assert(slashBond(10, 30,  Map("bob" -> Map(4 -> 10)), "alice", VALIDATORS.mapBy(v => if (v == "bob") [{epoch: 5, rate: 1}] else List())) == 20),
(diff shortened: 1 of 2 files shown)


2023/Q3/artifacts/PoS-quint/helper-functions.qnt
```diff
@@ -845,153 +872,133 @@ module helperFunctions {
...
-                                                                     val updatedBalanceRedelegatedBonds = computeBalanceRedelegatedBonds(totalRedelegatedBonded.get(e), e, curEpoch, slashes, totalRedelegatedUnbonded, acc._3)
-                                                                     val slashedAmount = (infractionStake - updatedTotalUnbonded) * finalRate
-                                                                     val currentStake = stakes.get(e) - acc._4.get(e)
-                                                                     val slashableStake = (currentStake - updatedBalanceBonds - updatedBalanceRedelegatedBonds.get(e))
-                                                                     (updatedTotalUnbonded, updatedBalanceBonds, updatedBalanceRedelegatedBonds, acc._4.set(e, acc._4.get(e) + min(slashedAmount, slashableStake))))
-      result._4
+      val bonds = totalBonded.keys().filter(e => e <= infractionEpoch).mapBy(start => totalBonded.get(start))
+      val redelegatedBonds = bonds.keys().mapBy(e => totalRedelegatedBonded.getOrElse(e, Map()))
+      val epochRange = (curEpoch+1).to(curEpoch+PIPELINE_OFFSET).sortSetDecreasing()
```
brentstone, line 873: previously this ranged from `curEpoch+1` to `curEpoch+PIPELINE_OFFSET+1`, why is it only to `curEpoch+PIPELINE_OFFSET` now?

2023/Q3/artifacts/PoS-quint/helper-functions.qnt
```diff
@@ -831,12 +831,39 @@ module helperFunctions {
...
+    // - @param start the epoch at which the bond started.
+    // - @param amount the bonded amount.
+    // - @param redelegatedBond a RedelegatedBondsMap that .
+    // - @param validator the address of the bonding validator.
+    // - @param slashes a map from address validator to a list of slashes.
+    // - @returns the amount of tokens left after slashing.
+    pure def computeBondAtEpoch(epoch: Epoch, start: Epoch, amount: int, redelegatedBond: RedelegatedBondsMap, validator: Address, slashes: Address -> List[Slash]): int = {
+       val listSlashes = slashes.get(validator).select(s => start <= s.epoch and s.epoch + slashProcessingDelay < epoch)
+       val x = printVariable("listSlashes", listSlashes)
+       val filteredSlashMap = slashes.keys().mapBy(v => slashes.get(v).select(s => s.epoch + slashProcessingDelay < epoch))
```
brentstone, line 844: Isn't this redundant with `val listSlashes` above?

2023/Q3/artifacts/PoS-quint/helper-functions.qnt
angbrav: Before it was also ranging from `curEpoch+1` to `curEpoch+PIPELINE_OFFSET`  as the interval generated by `range(curEpoch+1, curEpoch+PIPELINE_OFFSET+1)` is open on the high endpoint `curEpoch+PIPELINE_OFFSET+1`. See [Quint's syntax cheetsheet](https://github.com/informalsystems/quint/blob/main/doc/quint-cheatsheet.pdf).

2023/Q3/artifacts/PoS-quint/namada-redelegation.qnt
```diff
@@ -264,11 +268,7 @@ module namada {
...
-                                                                                                                                                                                    else 0))
-                                                                          .with("totalRedelegatedBonded", (curEpoch-CUBIC_OFFSET-UNBONDING_OFFSET+1).to(curEpoch+1+PIPELINE_OFFSET).mapBy(e => if (e < curEpoch+1+PIPELINE_OFFSET) validatorsState.get(v).totalRedelegatedBonded.get(e)
-                                                                                                                                                                                               else Map()))                                                                                                                                                                                    
+                                                                                                                                                                                                 else Map()))
```
brentstone, line 271: why remove these here? Seems like these lines are just clearing storage that is not going to be used again. What will the `totalBonded` and `totalRedelegatedBonded` entries still be used for possibly?

angbrav: Those lines do not make sense the way we use `totalBonded` and `totalRedelegatedBonded` in the current design. Before we were only maintaining `totalBonded` and `totalRedelegatedBonded` for an epoch window: from `curEpoch=CUBIC_OFFSET-UNBONDING_OFFSET` to `curEpoch+PIPELINE_OFFSET`, because those were the only entries required by the protocol (`slashValidator` function). Now, `totalBonded and totalRedelegatedBonded` are a strict fold over delegators of `bonded` and `redelegatedBonded` and we may need older entries in `slashValidator`.

PR #83, opened 2023-10-17
Rewards in Quint

angbrav:
This PR adds rewards to the Quint spec. The rewards system works as follows:
- Users have to submit a claim transaction for a given validator to claim all rewards accumulated since the last time they claimed rewards for that validator.
- Rewards are computed as follows:
  - At the end of an epoch, we compute for each validator a `rewardProduct`, which determines based on a list of slashes, the fraction of the validator's voting power at that epoch that should be produce rewards.
- Reward products are stored and applied both when unbonding or when claiming.
- Validators get a commission for each of the rewards.

Three aspects to bear in mind:
- The `rewardProducts` are computed before all slashes for a given validator are known. This may lead to produce greater rewards than it should be.
- The missing slashes may be processed in epoch between the epoch in which the reward is processed and the epoch in which the delegator claim its rewards. Even though those slashes were not considered when computing the `rewardProduct`, they are consider when computing the delegator's reward.
- Rewards are computed when unbonding. This implies that delegators do not receive rewards for `pipeline-offset` epochs: from the epoch they unbond until the unbond is materialized. This is because the delegator's tokens are still contributing to the validator's stake during those epochs.


2023/Q3/artifacts/PoS-quint/namada-redelegation.qnt
```diff
@@ -259,18 +293,20 @@ module namada {
...
-                                                                          .with("slashes", if (slashPerValidator.has(v)) validatorsState.get(v).slashes.append({epoch: infractionEpoch, rate: slashPerValidator.get(v)})
-                                                                                           else validatorsState.get(v).slashes))
+                                                                          .with("slashes", updatedSlashesMap.get(v))
+                                                                          .with("commission", (curEpoch-CUBIC_OFFSET-UNBONDING_OFFSET+1).to(curEpoch+1+PIPELINE_OFFSET).mapBy(e => if (e < curEpoch+1+PIPELINE_OFFSET) validatorsState.get(v).commission.get(e)
```
tzemanovic: what is happening with the commission rate here? Iiuc, we're copying the previous rates, except for epoch `curEpoch+1+PIPELINE_OFFSET` where it's set to `0`. Why is that?

angbrav: That's just wrong; fix coming..

angbrav:
Closed by https://github.com/informalsystems/partnership-heliax/pull/85

